PR #5, opened 2021-02-25
Hide plugins flag for now.

kku1993:


thatplguy (approved):

thatplguy:
Does this need to be removed from the `man` documentation too?

kku1993:
> Does this need to be removed from the `man` documentation too?

Yup, they are removed in this PR as well

PR #8, opened 2021-04-07
Retain the final timestamp for each request

mgritter:
Depends on https://github.com/akitasoftware/akita-libs/pull/12

Previously if a ParsedNetworkTraffic object spanned multiple packets, the timestamp on the first packet was recorded in ObservationTime. This change records the timestamp on the final packet in the sequence as well, which corresponds to the time when a request has completely arrived at a server.

Method signatures have changed to pass the reassembly.AssemblerContext through, which was previously ignored.


thatplguy (approved):
LGTM!

PR #9, opened 2021-04-08
Add daemon mode to expose some client functionality over REST

liujed:
This introduces a `daemon` command that exposes a REST API for client functionality. By default, the daemon listens on port 50080 and has the following endpoints.

* `POST /services/{serviceName}/learn/sessions`
  Creates a new learning session for the given service name.
* `POST /services/{serviceName}/learn/sessions/{learnSessionName}/witnesses`
  Adds a set of witnesses to a learning session.
  The request body is expected to be a stream of HAR entry objects to be added.
* `POST /services/{serviceName}/api-models`
  Creates an API model out of a set of learning sessions.
  The request body is expected to be a JSON object specifying the names of the learning sessions.


daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+type httpHandler func(*http.Request) HTTPResponse
+
+var cmdArgs Args
+
+func Run(args Args) error {
+	cmdArgs = args
+	router := mux.NewRouter().StrictSlash(true)
+
+	// Register an endpoint for creating a new learning session.
+	router.Handle("/services/{serviceName}/learn/sessions", httpHandler(createLearnSession)).Methods("POST")
```
mgritter, line 51: I am confused about how the pieces plug together here; how does mux know what to do with the HTTPResponse type that you defined?  Or the ClientResponse interface, which you also defined?

It's not a big deal, but if it works by satisfying some existing interface, I have found the idiom

var _ ClientResponse = &HttpResponse{}

to be useful for documenting the type assumption (whatever it is?) and giving the compiler a place to check the type assertion even if there's an interface{} or other non-type-checkable call involved.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+	Domain   string
+
+	// Optional args.
+	PortNumber uint16
+
+	Plugins []plugin.AkitaPlugin
+}
+
+// A wrapper around HTTP request handlers for handling HTTP responses generically.
+type httpHandler func(*http.Request) HTTPResponse
```
mgritter, line 42: I don't understand why the return type is HTTPResponse rather than *HTTPResponse (maybe it's to do with my confusion below about how it all works?)  It seems to require a lot of un-idiomatic pointer dereferencing, i.e, `*NewHTTPResponse(...)` instead of just `NewHttpResponse(...)`.

mgritter (approved):
Looks good, but did leave me a little confused.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+
+func Run(args Args) error {
+	cmdArgs = args
+	router := mux.NewRouter().StrictSlash(true)
+
+	// Register an endpoint for creating a new learning session.
+	router.Handle("/services/{serviceName}/learn/sessions", httpHandler(createLearnSession)).Methods("POST")
+
+	// Register an endpoint for adding witnesses to a learning session. The request body is expected to be a stream of HAR entry objects to be added.
+	router.Handle("/services/{serviceName}/learn/sessions/{learnSessionName}/witnesses", httpHandler(addWitnesses)).Methods("POST")
```
thatplguy, line 62: Can you preface each of the paths with `/v1`, e.g. `/v1/services/...`?  We've adopted that style of versioning elsewhere, and it will make it easier to introduce new versions in the future.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// Register an endpoint for creating an API model out of a set of learning sessions. The request body is expected to be a JSON object specifying the names of the learning sessions.
```
thatplguy, line 64: Nit: How do you feel about wrapping comments at e.g. 100 characters?  AFAIK it's not part of Go's style guidelines, but I've found it makes reading PRs easier.  No change required for this PR, unless you feel inspired to do it :)

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+	vars := mux.Vars(request)
+
+	// Get the service ID.
+	serviceID, httpErr := getServiceID(vars)
+	if httpErr != nil {
+		return *httpErr
+	}
+
+	// Create a new learning session with a random name.
+	// TODO: Allow user to specify a name?
```
thatplguy, line 95: Yes, this would be nice to add down the road.  Not necessary for this PR.

daemon/run.go
liujed: Addressed in 77443df.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+	// Start a consumer goroutine that will read witnesses from a channel and pass them on to a backend collector.
+	witnessChannel := make(chan *Witness, WITNESS_BUFFER_SIZE)
+	doneChannel := make(chan bool)
+	go func() {
+		defer func() { doneChannel <- true }()
+
+		// Create collector for ingesting the witnesses.
+		collector := trace.NewBackendCollector(serviceID, learnSessionID, learnClient, kgxapi.Inbound, cmdArgs.Plugins)
+
+		// TODO: Offer option to remove third-party trackers?
```
thatplguy, line 150: Yes, this would be great.  I believe there's logic to do this already in the CLI.  Can we factor it out so it's uniformly applied?

Not necessary for this PR, but necessary before we release this.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	witDetails := witnessDetails{
+		Parsed:    numWitnessesParsed,
+		Drops:     numWitnessesDropped,
+		Errors:    sampledErrs.TotalCount,
+		Successes: successfulEntries,
+	}
+
+	// If we encountered malformed witnesses or there were errors processing witnesses, return a "Bad Request" status.
```
thatplguy, line 235: It would be good to distinguish between parsing errors (400 response) throttling errors (429 response) and internal errors (5xx response).  Throttling is particularly important, because the client can use that information to alter the sampling rate.

daemon/run.go
```diff
@@ -0,0 +1,319 @@
...
+	for i, learnSessionName := range requestBody.LearnSessionNames {
+		var err error
+		learnSessionIDs[i], err = util.GetLearnSessionIDByName(learnClient, learnSessionName)
+		if err != nil {
+			return *NewHTTPError(err, http.StatusNotFound, "Learning session not found: "+learnSessionName)
+		}
+	}
+
+	// Create a new API model with a random name.
+	// TODO: Allow user to specify a name?
```
thatplguy, line 301: Also good to have, but not necessary for this PR.

thatplguy (changes requested):
Looks great, modulo a few comments!  The only required change in this PR is to preface all endpoints with `/v1`.  Other changes, like exposing a throttled response code, are nice to have but can be in future PRs.

daemon/run.go
liujed: The return type is `HTTPResponse` because I want to force implementations to provide an actual response; they shouldn't be able to return `nil`. In 638cf0a, I've tweaked things so we have fewer pointer dereferences now.

daemon/run.go
liujed: Done in 2ed6419.

daemon/run.go
liujed: I'll have to figure out how to convince VSCode to wrap just comments and nothing else.

daemon/run.go
liujed: Indeed, the logic is there. The question is, how do we want it applied? Per request, or per daemon instance (i.e., specified in each request or on the command line)?

daemon/run.go
liujed: Here are the differences between what you propose and what is implemented:
* Errors returned from `apispec.ProcessHAREntry` are lumped in with parsing errors and yield a 400 status.
* Throttling errors currently yield a 422 status.

I can make the 422→429 change. It wasn't clear to me that errors from `apispec.ProcessHAREntry` are internal errors. But if that's what they are, I can have them produce a 500 status.

daemon/run.go
thatplguy: On learn session creation, maybe?  That gives the client the option to change things without redeploying (e.g. by changing config options in the environment), but it doesn't add an argument that's unlikely to change to every witness request.

liujed: This would introduce state that we currently don't keep. The dæmon itself would need to remember whether tracker removal is desired, and this state would need to persist across restarts of the dæmon. My vote is for the flag to be provided at the start of each `addWitnesses` call. I'm expecting there to be multiple witnesses for each call, which would amortize the overhead.

daemon/run.go
thatplguy: `apispec.ProcessHAREntry` can encounter different kinds of errors, although it's not set up right now to expose them in an actionable way.

The first kind are errors returned from the backend.  This is where throttling errors would come from, as well as (hopefully transient) 5xx errors.

The second are errors `apispec.ProcessHAREntry` itself generates, which are hopefully just validation errors.

Ideally, we would be able to distinguish between different kinds of errors and pass them back to client.  (Best practice in API design is to pass errors back, rather than e.g. doing retries yourself in the face of throttling.)

For now, using `422` as a catchall is fine.  We can add handling for throttling and other specific errors later, after the daemon and middleware are done.

daemon/run.go
thatplguy: Sounds good to me.

daemon/run.go
mgritter: Got it, thanks, didn't see the ServeHTTP function.

daemon/run.go
liujed: Done in b2ec1c3.

daemon/run.go
liujed: Done in 25b4986.

daemon/run.go
liujed: I'll leave this as is. We can improve this when `apispec.ProcessHAREntry` has finer grained error reporting.

daemon/run.go
liujed: Done in 2231328.

daemon/run.go
liujed: Done in 2231328.

thatplguy (approved):
LGTM!

liujed:
This introduces a `daemon` command that exposes a REST API for client functionality. By default, the daemon listens on port 50080 and has the following endpoints.

* `POST /services/{serviceName}/learn/sessions`
  Creates a new learning session for the given service name.
* `POST /services/{serviceName}/learn/sessions/{learnSessionName}/witnesses`
  Adds a set of witnesses to a learning session.
  The request body is expected to be a stream of HAR entry objects to be added.
* `POST /services/{serviceName}/api-models`
  Creates an API model out of a set of learning sessions.
  The request body is expected to be a JSON object specifying the names of the learning sessions.

PR #10, opened 2021-04-08
Use ParsedNetworkTraffic timestamps for HAL generation

mgritter:
With this change, the HAR file now looks like:

```
      {
        "_id": "wit_2o9OnzUJzu44TmxnwUGMYt",
        "startedDateTime": "2021-04-08T09:57:04.639398-07:00",
        "time": 1002.753,
        "request": ...
        ...
        "timings": {
          "send": 0,
          "wait": 1002.753,
          "receive": 0
        }
      },
```

(Send and receive are zero in this example from akibix because they are single packets so we don't have any duration associated with them.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,5 +40,5 @@ require (
 
 replace (
 	github.com/google/gopacket v1.1.18 => github.com/akitasoftware/gopacket v1.1.18-0.20201119235945-f584f5125293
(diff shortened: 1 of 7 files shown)


mgritter:
Depends on changes to akitasoftware/martian

thatplguy (approved):

PR #12, opened 2021-04-09
Put the processing latency measurement into the IR

mgritter:
Calculate the latency from the information in the ParsedNetworkTraffic objects.

For HAR replay, rebuild ParsedNetworkTraffic timestamps so that they will give
the same latency result (as is already present in the HAR's Timings.Wait. This is sort of gross, but
there didn't seem to be a good way to get the information from point A to point B without more
work than it was worth.

The original code had comments suggesting that type incompatibility prevented re-use 
of the Martian library's HAR parsing; I'm guessing that is because it originally used ints?


mgritter:
Marked as draft because it has to go in *after* changes to the server side, but tested and appears to work.

thatplguy (approved):
Looks great!

apispec/har.go
```diff
@@ -39,6 +41,67 @@ func processHAR(inboundCol, outboundCol trace.Collector, p string) error {
...
+	ResponseStart time.Time
+	ResponseEnd   time.Time
+}
+
+func advanceTime(current time.Time, ms *float32) time.Time {
+	if ms == nil || *ms <= 0.0 {
+		return current
+	}
+	// convert ms to nanoseconds
+	duration := time.Duration(int64(*ms * 1_000_000.0))
```
liujed, line 65: How about `duration := *ms * time.Millisecond`?

trace/backend_collector.go
```diff
@@ -73,6 +76,39 @@ func (r witnessWithInfo) toReport(dir kgxapi.NetworkDirection) (*kgxapi.WitnessR
...
+
+	// Missing data, leave as default value in protobuf
+	if requestEnd.IsZero() || responseStart.IsZero() {
+		return
+	}
+
+	// HTTPMethodMetadata only for now
+	if meta := spec_util.HTTPMetaFromMethod(w.witness.Method); meta != nil {
+		latency := responseStart.Sub(requestEnd)
+		meta.ProcessingLatency = float32(latency.Microseconds()) / 1000.0
```
liujed, line 108: How about `meta.ProcessingLatency = float32(latency/time.Millisecond)`?

liujed (approved):

apispec/har.go
mgritter: That's where I started, it doesn't work.  Only constants can do that.  :(

https://play.golang.org/p/f71AyeNmJoG

./prog.go:9:18: invalid operation: *ms * time.Millisecond (mismatched types float32 and time.Duration)

trace/backend_collector.go
mgritter: Same problem, can't divide a float by a Duration, have to convert it to a Duration first.  (And the time you get is then a Duration, which is an int64, so the fractional part would be gone.)

liujed: `latency / time.Millisecond` does type-check, but you're right that you lose the fractional part. https://play.golang.org/p/kTyRO1PFp0b

apispec/har.go
liujed: Alas, how disappointing.

PR #13, opened 2021-04-15
Add support for uploading HAR files to `upload`

liujed:
This adds support for uploading HAR files to the `upload` command. The old `--service` and `--name` options are now deprecated in favour of `--dest`, which takes in Akita URIs.

When uploading traces, the user may provide multiple files to upload to a single trace. Plugins and removal of third-party trackers is supported. If the specified trace already exists, the user must specify `--append` to modify the trace. If `--append` is used when the trace doesn't yet exist, the trace is created and a warning is printed to stderr.

### DIFF
--- a/cmd/internal/upload/cmd.go
+++ b/cmd/internal/upload/cmd.go
@@ -1,9 +1,11 @@
 package upload
 
 import (
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/akitasoftware/akita-libs/akid"
+	"github.com/akitasoftware/akita-libs/akiuri"
 
 	"github.com/akitasoftware/akita-cli/cmd/internal/akiflag"
 	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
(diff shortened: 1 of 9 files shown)


liujed:
[ch2768]

cmd/internal/upload/cmd.go
```diff
@@ -1,27 +1,78 @@
...
+		// Rewrite --service and --name into --dest.
+		if destFlag == "" {
+			destFlag = "akita://" + serviceFlag + ":spec"
+			if specNameFlag != "" {
+				destFlag += ":" + specNameFlag
+			}
+		}
+
+		// Parse --dest.
+		destURI, err := akiuri.Parse(destFlag)
```
mgritter, line 45: I have had trouble with this when the specName is missing:

mark@ubuntu:~/akibox-tutorial$ sudo ../akita-cli/bin/akita --domain 127.0.0.1:50443 --test_only_disable_https learn --port 8000 --out akita://akibox:trace
[ERROR] must specify --service when --out is not an AkitaURI

So I *think* this does not work to automatically generate a trace name.

mgritter (approved):
Looks good, but I have a concern about the akita URI with a missing trace name.

cmd/internal/upload/cmd.go
liujed: It doesn't for `learn`, but `upload` generates a trace name if you say `--dest akita://akibox:trace:`.

thatplguy (approved):
LGTM once the URI parsing allows for `akita://service:kind` without the trailing `:`.

PR #16, opened 2021-04-16
Improvements and fixes to AkitaURIs

liujed:
We can now use Akita URIs without object names for the commands `apidump`, `apispec`, and `apilearn`. For example, `akita learn --port 8000 --out akita://akibox-rails:spec`.

For convenience, the object type can also be omitted in contexts where the object type is unambiguous. For example, `akita learn --port 8000 --out akita://akibox-rails`.

Fixed handling of `--service` and `--out` for `apidump`: exactly one should be specified.

Error out early if a non-trace URI is provided to `apidump --out`.

In `apispec` and `learn`, error out if `--service` and `--out` refer to different services.

In `learn`, error out early if `--out` refers to a spec that already exists or is given a non-spec URI.

Depends on akitasoftware/akita-libs#18.


mgritter (approved):
LGTM.

PR #17, opened 2021-04-23
Resolve race conditions in BackendCollector.

mgritter:
Replace map[] with a sync.Map.
Move periodic flush to its own goroutine.

The unit test is not too impressive right now but I intend to use it to test packet count per interface, when that goes in.

### DIFF
--- a/trace/backend_collector.go
+++ b/trace/backend_collector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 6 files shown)


liujed (approved):
Looks reasonable to me!

trace/backend_collector.go
```diff
@@ -252,11 +247,27 @@ func (c *BackendCollector) uploadWitnesses(in []interface{}) {
 	}
 }
 
+func (c *BackendCollector) periodicFlush() {
```
thatplguy, line 250: Where does this get called?

mgritter: Good catch, I had intended to call it in NewBackendCollector as a goroutine.

mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/17/commits/f88f0281d685bc49f4daa35458ccd61036862fde

thatplguy (approved):
LGTM!

PR #18, opened 2021-04-26
Diagnostic information when trace is empty

mgritter:
Capture packet counts from pcap and parsing layers.
Display this information in summary form in the debug output.

Example output capturing HTTPS:
```
[INFO] Captured 364 TCP packets total; 182 unparsed TCP segments.
[INFO] This may mean you are trying to capture HTTPS traffic.
[INFO] See https://docs.akita.software/docs/proxy-for-encrypted-traffic
[INFO] for instructions on using a proxy, or generate a HAR file with
[INFO] your browser as described in
[INFO] https://docs.akita.software/docs/collect-client-side-traffic-2
[ERROR] No inbound HTTP calls captured! 🛑

[ERROR] failed to create trace: incomping API trace is empty
```

Example output when no traffic on the desired port:
```
[INFO] Did not capture any TCP packets matching the filter.
[INFO] This may mean your filter is incorrect, such as the wrong TCP port.
[ERROR] No inbound HTTP calls captured! 🛑

[ERROR] failed to create trace: incomping API trace is empty
```

Example debug output:
```
[DEBUG] ==================================
[DEBUG] Packets per interface:
[DEBUG]       interface      dir packets   req  resp   unk
[DEBUG]              lo  INBOUND       0     0     0     0
[DEBUG]              lo OUTBOUND       0     0     0     4
[DEBUG]           ens33  INBOUND       0     0     0     0
[DEBUG]           ens33 OUTBOUND       1     0     0    12
[DEBUG]         docker0  INBOUND       0     0     0     0
[DEBUG]         docker0 OUTBOUND       0     0     0     0
[DEBUG] br-25f3702c5d47  INBOUND       0     0     0     0
[DEBUG] br-25f3702c5d47 OUTBOUND       0     0     0     0
[DEBUG]     veth78430a1  INBOUND       0     0     0     0
[DEBUG]     veth78430a1 OUTBOUND       0     0     0     0
[DEBUG] ==================================
[DEBUG] Packets per port:
[DEBUG]     port packets   req  resp   unk
[DEBUG] --------- matching filter --------
[DEBUG]        no packets captured        
[DEBUG] ------- not matching filter ------
[DEBUG]    40005       0     0     0     2
[DEBUG]    44790       1     0     0     0
[DEBUG]    49513       0     0     0     2
[DEBUG]       53       0     0     0    12
[DEBUG]    44854       0     0     0     2
[DEBUG]    54092       0     0     0     4
[DEBUG]     1900       0     0     0     4
[DEBUG]      443       1     0     0     0
[DEBUG]    53897       0     0     0     2
[DEBUG]    37949       0     0     0     4
[DEBUG] ==================================
```

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -76,6 +76,9 @@ type Args struct {
 	Plugins []plugin.AkitaPlugin
 }
 
+// DumpPacketCounters prints the accumulated packet counts per interface and per port,
+// at Debug level, to stderr.  The first argument should be the keyed by interface names (as created
+// in the Run function below); all we really need are those names.
 func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *trace.PacketCountSummary, outboundSummary *trace.PacketCountSummary) {
 	// Using a map gives inconsistent order when iterating (even on the same run!)
 	directions := []kgxapi.NetworkDirection{kgxapi.Inbound, kgxapi.Outbound}
@@ -84,8 +87,7 @@ func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *tra
 		toReport = append(toReport, outboundSummary)
 	}
 
-	// Debugging info: packets per interface and prot
-	printer.Stderr.Debugf("==================================\n")
+	printer.Stderr.Debugf("===============================================\n")
 	printer.Stderr.Debugf("Packets per interface:\n")
 	printer.Stderr.Debugf("%15v %8v %7v %5v %5v %5v\n", "interface", "dir", "packets", "req", "resp", "unk")
 	for n := range interfaces {
@@ -102,7 +104,7 @@ func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *tra
 		}
 	}
 
-	printer.Stderr.Debugf("==================================\n")
+	printer.Stderr.Debugf("===============================================\n")
 	printer.Stderr.Debugf("Packets per port:\n")
 	printer.Stderr.Debugf("%8v %7v %5v %5v %5v\n", "port", "packets", "req", "resp", "unk")
 	for i, summary := range toReport {
@@ -128,7 +130,8 @@ func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *tra
 		}
 	}
 
-	printer.Stderr.Debugf("==================================\n")
+	printer.Stderr.Debugf("===============================================\n")
+
 }
 
 func Run(args Args) error {
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -169,6 +169,9 @@ func Run(args Args) error {
 		}
 	}
 
+	// Initialize packet counts
+	summary := trace.NewPacketCountSummary()
+
 	// Start collecting
 	var doneWG sync.WaitGroup
 	doneWG.Add(len(inboundFilters) + len(outboundFilters))
@@ -196,11 +199,11 @@ func Run(args Args) error {
 
 				if args.Out.AkitaURI != nil && args.Out.LocalPath != nil {
 					collector = trace.TeeCollector{
-						Dst1: trace.NewBackendCollector(backendSvc, backendLrn, learnClient, dir, args.Plugins),
+						Dst1: trace.NewBackendCollector(backendSvc, backendLrn, learnClient, dir, args.Plugins, summary),
 						Dst2: localCollector,
 					}
 				} else if args.Out.AkitaURI != nil {
-					collector = trace.NewBackendCollector(backendSvc, backendLrn, learnClient, dir, args.Plugins)
+					collector = trace.NewBackendCollector(backendSvc, backendLrn, learnClient, dir, args.Plugins, summary)
 				} else if args.Out.LocalPath != nil {
 					collector = localCollector
 				} else {
@@ -220,7 +223,7 @@ func Run(args Args) error {
 			go func(interfaceName, filter string) {
 				defer doneWG.Done()
 				// Collect trace. This blocks until stop is closed or an error occurs.
-				if err := trace.Collect(stop, interfaceName, filter, collector); err != nil {
+				if err := trace.Collect(stop, interfaceName, filter, collector, summary); err != nil {
 					errChan <- errors.Wrapf(err, "failed to collect trace on interface %s", interfaceName)
 				}
 			}(interfaceName, filter)
@@ -306,6 +309,19 @@ func Run(args Args) error {
 	if stopErr != nil {
 		return errors.Wrap(stopErr, "trace collection failed")
 	}
+
+	printer.Stderr.Infof("%15v %7v %5v %5v %5v\n", "interface", "packets", "req", "resp", "unk")
+	for n := range interfaces {
+		count := summary.TotalOnInterface(n)
+		printer.Stderr.Infof("%15s %7d %5d %5d %5d\n",
+			n,
+			count.TCPPackets,
+			count.HTTPRequests,
+			count.HTTPResponses,
+			count.Unparsed,
+		)
+	}
+
 	printer.Stderr.Infof("%s 🎉\n\n", aurora.Green("Success!"))
 
 	return nil
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -400,7 +400,7 @@ func Run(args Args) error {
 			printer.Stderr.Infof( "https://docs.akita.software/docs/collect-client-side-traffic-2\n" )
 		}
 		printer.Stderr.Errorf("%s 🛑\n\n", aurora.Red("No inbound HTTP calls captured!"))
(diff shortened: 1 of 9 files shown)


mgritter:
Depends on https://github.com/akitasoftware/akita-libs/pull/20 to compile.

apidump/apidump.go
```diff
@@ -76,6 +76,61 @@ type Args struct {
...
+	directions := []kgxapi.NetworkDirection{kgxapi.Inbound, kgxapi.Outbound}
+	toReport := []*trace.PacketCountSummary{inboundSummary}
+	if outboundSummary != nil {
+		toReport = append(toReport, outboundSummary)
+	}
+
+	// Debugging info: packets per interface and prot
+	printer.Stderr.Debugf("==================================\n")
+	printer.Stderr.Debugf("Packets per interface:\n")
+	printer.Stderr.Debugf("%15v %8v %7v %5v %5v %5v\n", "interface", "dir", "packets", "req", "resp", "unk")
```
liujed, line 90: Why not write out "requests", "responses", and "unknown" in full? Similarly elsewhere.

apidump/apidump.go
```diff
@@ -76,6 +76,61 @@ type Args struct {
...
+func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *trace.PacketCountSummary, outboundSummary *trace.PacketCountSummary) {
+	// Using a map gives inconsistent order when iterating (even on the same run!)
+	directions := []kgxapi.NetworkDirection{kgxapi.Inbound, kgxapi.Outbound}
+	toReport := []*trace.PacketCountSummary{inboundSummary}
+	if outboundSummary != nil {
+		toReport = append(toReport, outboundSummary)
+	}
+
+	// Debugging info: packets per interface and prot
+	printer.Stderr.Debugf("==================================\n")
```
liujed, line 88: Make this span the whole width of the table? Similarly elsewhere.

pcap/net_parse.go
```diff
@@ -48,18 +48,27 @@ func (fact *tcpStreamFactory) New(netFlow, tcpFlow gopacket.Flow, _ *layers.TCP,
 	return newTCPStream(fact.clock, netFlow, fact.outChan, fact.fs)
 }
 
+type NetworkTrafficObserver func(gopacket.Packet)
```
liujed, line 51: Please document.

pcap/net_parse.go
```diff
@@ -48,18 +48,27 @@ func (fact *tcpStreamFactory) New(netFlow, tcpFlow gopacket.Flow, _ *layers.TCP,
...
+
 type NetworkTrafficParser struct {
 	pcap  pcapWrapper
 	clock clockWrapper
+
+	observer NetworkTrafficObserver
```
liujed, line 57: Please document the purpose of this field.

trace/run.go
```diff
@@ -8,20 +8,27 @@ import (
...
 	facts := []akinet.TCPParserFactory{
 		akihttp.NewHTTPRequestParserFactory(),
 		akihttp.NewHTTPResponseParserFactory(),
 	}
+
 	parser := col.NewNetworkTrafficParser()
+
+	if packetCount != nil {
+		parser.InstallObserver(CountTcpPackets(intf, packetCount))
+	}
```
liujed, line 23: Why not roll this into the constructor?

trace/run.go
```diff
@@ -8,20 +8,27 @@ import (
...
+
 	parsedChan, err := parser.ParseFromInterface(intf, bpfFilter, stop, facts...)
 	if err != nil {
 		return errors.Wrap(err, "couldn't start parsing from interface")
 	}
 
 	for t := range parsedChan {
+		t.Interface = intf
```
liujed, line 31: Why not fold this into `ParseFromInterface`?

trace/stats.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+		new.Add(c)
+		s.byPort[new.SrcPort] = new
+	}
+
+	if prev, ok := s.byPort[c.DstPort]; ok {
+		prev.Add(c)
+	} else {
+		// Use SrcPort as the identifier in the
+		// accumualted counter
```
liujed, line 92: Is the SrcPort/DstPort distinction used anywhere?

Also, typo: accumulated.

trace/stats.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// A consumer accepts incremental updates in the form
+// of PacketCounters.
+type PacketCountConsumer interface {
+	// Add an additional measurement to the current count
+	Update(delta PacketCounters)
+
+	// Signal the end of the measurement interval
+	Close()
```
liujed, line 40: Does this have any non-empty implementations?

liujed (approved):
A pile of nits, but looks good.

trace/run.go
mgritter: I wasn't sure how many times NewNetworkTrafficParser was being used.  (The answer is it's only 4, so it would have been reasonable.)

The idiomatic issue I have is that I hate adding nil's (even though I did it elsewhere.)  It is much better, in my opinion, to have either

```
newObj := Blah()
newObj.SetUpFoo()
```

or

newObj := Blah( &BlahConfig {
  Foo: fooConfig,
} )
```

than

```
newObj := Blah( nil, nil, fooConfig, nil )
```

But, I didn't have the guts to refactor NewBackendCollector into the latter form.

File: trace/run.go (hunk 6 above)
Comment (mgritter) [reply trace/run.go:31]: That makes a lot of sense, but I could not make it work easily.  Some ParsedNetworkPackets are generated directly in parser, but the ones we care about come from TCP reassembly, so the interface name would have to be added to that context so that the callback which eventually creates the HTTP Response could find it.

I think the appropriate place would be 
```
type assemblerCtxWithSeq struct {
	ci       gopacket.CaptureInfo
	seq, ack reassembly.Sequence
}
```

File: trace/stats.go (hunk 8 above)
Comment (mgritter) [reply trace/stats.go:40]: No, I'll remove it.

My original thinking was to use a worker goroutine to sum everything instead of doing it in a mutex-protected critical section. so I needed a Close to shut that down.  But I don't think I actually called Close so it wouldn't work anyway.

Comment (mgritter) [reply trace/stats.go]: Fixed in https://github.com/akitasoftware/akita-cli/pull/18/commits/5b903c0c1e8b5884c0cb3595f8cf212de2ae0870

File: trace/run.go (hunk 6 above)
Comment (mgritter) [reply trace/run.go:31]: Maybe ParseFromInterface could have a separate goroutine to re-stamp every ParsedNetworkPacket with the right interface, and put it on a different channel?  Still a bunch of work to get right.

File: trace/stats.go (hunk 7 above)
Comment (mgritter) [reply trace/stats.go:86]: The original data has SrcPort and DstPort filled in.

The problem is that I couldn't think of a good way to summarize all the port-pairs compactly, so this implementation doesn't track them.  (It might be that the right thing is to collect the entire 4-tuple so we can track packets per flow?)  But I think they should be filled in correctly at the source of data in case we figure out how to do better.

File: apidump/apidump.go (hunk 1 above)
Comment (mgritter) [reply apidump/apidump.go:90]: 🤷 I find it easier to read

```
[DEBUG]       interface      dir packets   req  resp   unk
[DEBUG]              lo  INBOUND       0     0     0     0
[DEBUG]              lo OUTBOUND       0     0     0     4
```

than
```
[DEBUG]       interface      dir packets requests responses unknown
[DEBUG]              lo  INBOUND       0        0         0       0
[DEBUG]              lo OUTBOUND       0        0         0       4
```

since I thought 99999 was a reasonable (but non-enforced) upper bound on the number of packets.  I thought about going down to 4 characters per column :)

File: pcap/net_parse.go (hunk 4 above)
Comment (mgritter) [reply pcap/net_parse.go]: Fixed.

File: pcap/net_parse.go (hunk 3 above)
Comment (mgritter) [reply pcap/net_parse.go:54]: Fixed in https://github.com/akitasoftware/akita-cli/pull/18/commits/956c721b9bd0c25cc7d45c5583260cf4c5a70f3c

File: apidump/apidump.go (hunk 2 above)
Comment (mgritter) [reply apidump/apidump.go]: Fixed in https://github.com/akitasoftware/akita-cli/pull/18/commits/956c721b9bd0c25cc7d45c5583260cf4c5a70f3c

(It spans the width of the narrower table in its original form; I wanted all three the same, so now they're all longer.)

File: apidump/apidump.go [hunk 9]
```diff
@@ -76,6 +76,61 @@ type Args struct {
 	Plugins []plugin.AkitaPlugin
 }
 
+func DumpPacketCounters(interfaces map[string]interfaceInfo, inboundSummary *trace.PacketCountSummary, outboundSummary *trace.PacketCountSummary) {
```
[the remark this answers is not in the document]
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/18/commits/956c721b9bd0c25cc7d45c5583260cf4c5a70f3c

PR #20, opened 2021-04-30
BackendCollectors should send all requests to server

mgritter:
In-memory buffer is flushed on Close(), or periodically, so < 10 packets
wouldn't have been sent at all-- and larger traces would probably be truncated.

Added debugging information about flushes, and a count of uploaded HTTP packets.


mgritter:
How can we write a test for this?

liujed (approved):
Thanks. I think I modelled this code on `apidump`. It's not obvious where the call to `Close` there is happening. (Turns out it's handled by `trace.Collect`.)

PR #21, opened 2021-04-30
Rewrote the daemon

liujed:
This version of the daemon provides two API endpoints:
* `/v1/services/{serviceName}/middleware` for polling for changes in the active set of traces for a given service, and
* `/v1/services/{serviceName}/traces/{traceName}/events` for adding events to a trace.

Depends on akitasoftware/akita-libs#23 and akitasoftware/superstar#248.


cmd/internal/daemon/cmd.go
```diff
@@ -44,6 +48,14 @@ var Cmd = &cobra.Command{
...
 
 func init() {
+	Cmd.Flags().StringVar(
+		&nameFlag,
+		"name",
+		"",
+		"The name of the daemon. Used to identify this daemon in Akita Cloud.",
+	)
+	cobra.MarkFlagRequired(Cmd.Flags(), "name")
+
```
mgritter, line 58: I wonder if it makes sense using an automatically-generated name like we do for everything else.  I'd certainly prefer that users gave the daemon a meaningful name, but it might be a roadbump to force them to do so.

daemon/internal/cloud_client/change_active_traces_event.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Specifies which traces have been activated, and which have been
+	// deactivated.
+	activeTraceDiff daemon.ActiveTraceDiff
+}
+
+func newChangeActiveTracesEvent(serviceID akid.ServiceID, activeTraceDiff daemon.ActiveTraceDiff) changedActiveTracesEvent {
+	return changedActiveTracesEvent{
+		serviceID:       serviceID,
+		activeTraceDiff: activeTraceDiff,
```
mgritter, line 24: It feels sort of like activeTraceDiff should be passed around by pointer here (and elsewhere) instead of copied, but since it's just a collection of slices it's probably fine?

daemon/internal/cloud_client/change_active_traces_event.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func (event changedActiveTracesEvent) handle(client *cloudClient) {
+	printer.Debugf("Handling changed-traces event for service %s\n", akid.String(event.serviceID))
+
+	serviceInfo, ok := client.serviceInfoByID[event.serviceID]
+	if !ok {
+		printer.Debugf("Ignoring diff for unknown service: %s\n", akid.String(event.serviceID))
```
mgritter, line 33: If this shouldn't ever happen, then perhaps the log message should be Warning.

daemon/internal/cloud_client/client_state.go
```diff
@@ -0,0 +1,76 @@
+package cloud_client
+
+import (
+	"github.com/akitasoftware/akita-cli/rest"
+	"github.com/akitasoftware/akita-libs/akid"
+	"github.com/akitasoftware/akita-libs/daemon"
+)
+
+// Logging state for a single service.
```
mgritter, line 9: It would have been helpful, for my understanding, to have a comment here along the lines "all the modifications to this are done by a single worker goroutine" because I started thinking about race conditions.

I was worried about the large number of events making modifications to serviceInfo state, and was going to suggest they be moved into their own functions for encapsulation purposes, but I think I understand the structure better now.

daemon/internal/cloud_client/cloud_client.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Start the main goroutine for the cloud client.
+	//
+	// Accesses to anything inside client.serviceInfoByID must be done in this
+	// goroutine.
+	go func() {
+		for event := range client.eventChannel {
+			event.handle(client)
+		}
+	}()
```
mgritter, line 57: Debug message for main worker shutdown?

daemon/internal/cloud_client/cloud_client.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	return result
+}
+
+// Starts a goroutine for collecting trace events and sending them to the
+// cloud. Assumes the given service ID has been registered.
+func (client *cloudClient) startTraceEventCollector(serviceID akid.ServiceID, loggingOptions daemon.LoggingOptions) {
+	serviceInfo, traceInfo := client.getInfo(serviceID, loggingOptions.TraceID)
+	if serviceInfo == nil {
+		printer.Debugf("Got a new trace from the cloud for an unregistered service: %q\n", akid.String(serviceID))
```
mgritter, line 113: Warning?

daemon/internal/cloud_client/cloud_client.go
```diff
@@ -0,0 +1,199 @@
...
+	learnClient := serviceInfo.learnClient
+	filterThirdPartyTrackers := loggingOptions.FilterThirdPartyTrackers
+	traceEventChannel := make(chan *TraceEvent, TRACE_BUFFER_SIZE)
+	go func() {
+		// Create the collector.
+		collector := trace.NewBackendCollector(serviceID, loggingOptions.TraceID, learnClient, api_schema.Inbound, client.plugins, nil)
+		if filterThirdPartyTrackers {
+			collector = trace.New3PTrackerFilterCollector(collector)
+		}
+		defer collector.Close()
```
mgritter, line 138: This goroutine defined inline is too big, can it be feasibly made into its own function?

daemon/internal/cloud_client/cloud_client.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+
+			// Pass the trace event to the collector.
+			if entrySuccess := apispec.ProcessHAREntry(collector, streamID, seqNum, *traceEvent, &sampledErrs); entrySuccess {
+				successfulEntries += 1
+			}
+		}
+
+		// Log successfulEntries and sampledErrs.
+		printer.Debugln("Collected %d entries for trace %q\n", successfulEntries, loggingOptions.TraceID)
```
mgritter, line 159: Info?

daemon/internal/cloud_client/trace_event_request.go
```diff
@@ -0,0 +1,215 @@
...
+		// Log any errors that we encountered while processing the trace events.
+		eventDetails := TraceEventDetails{
+			Parsed: numTraceEventsParsed,
+			Drops:  numTraceEventsDropped,
+		}
+
+		// Send the result to the client.
+		{
+			defer close(req.responseChannel)
+
```
mgritter, line 156: I don't know if you are expecting the defer to exit at the end of the block, but it does not.  (If not, I don't understand why this section is in a curly-brace block.)

daemon/internal/cloud_client/long_poll_service_event.go
```diff
@@ -0,0 +1,43 @@
...
+	currentTraces := client.getCurrentTraces(event.serviceID)
+	frontClient := client.frontClient
+	go func() {
+		// Send a request to the cloud containing a list of the traces currently
+		// being logged. The response will be a list of new traces to log and a
+		// list of traces that should finish logging.
+		activeTraceDiff, err := util.LongPollActiveTracesForService(frontClient, event.serviceID, currentTraces)
+		if err != nil {
+			// Log the error, wait a bit, and try again.
+			printer.Debugf("Error while polling service ID %s: %v\n", akid.String(event.serviceID), err)
```
mgritter, line 33: Maybe Warning or Info.

daemon/internal/cloud_client/long_poll_service_event.go
```diff
@@ -0,0 +1,43 @@
...
+			time.Sleep(LONG_POLL_INTERVAL)
+			client.eventChannel <- event
+			return
+		}
+
+		// Enqueue a changeActiveTracesEvent for the main goroutine to handle.
+		printer.Debugf("Enqueuing changed-traces event for service %s\n", akid.String(event.serviceID))
+		client.eventChannel <- newChangeActiveTracesEvent(event.serviceID, activeTraceDiff)
```
mgritter, line 41: I find it kind of subtle that newChangeActiveTracesEvent is responsible for re-enqueing the long poll.  Is there a reason not to do it immediately?

daemon/internal/cloud_client/change_active_traces_event.go
```diff
@@ -0,0 +1,77 @@
...
+	// will be against the newly updated state.
+	channelsToSend := serviceInfo.responseChannels
+	serviceInfo.responseChannels = []chan<- daemon.ActiveTraceDiff{}
+
+	// Start a bunch of goroutines to send our responses.
+	for _, responseChannel := range channelsToSend {
+		go func(responseChannel chan<- daemon.ActiveTraceDiff) {
+			defer close(responseChannel)
+			responseChannel <- event.activeTraceDiff
+		}(responseChannel)
```
mgritter, line 72: If you just put these all on the channel immediately, they would be guaranteed to be handled in-order, I think.

By doing it in a goroutine now I have to wonder about if some of them could be indefinitely deferred, and get weird out-of-order operation on your state machine. I don't think there's a specific problem here but it's not within my current capability to say for sure.

daemon/internal/cloud_client/cloud_client.go
```diff
@@ -0,0 +1,199 @@
...
+func (client *cloudClient) getCurrentTraces(serviceID akid.ServiceID) []akid.LearnSessionID {
+	result := []akid.LearnSessionID{}
+	for traceID := range client.serviceInfoByID[serviceID].traces {
+		result = append(result, traceID)
+	}
+	return result
+}
+
+// Starts a goroutine for collecting trace events and sending them to the
+// cloud. Assumes the given service ID has been registered.
```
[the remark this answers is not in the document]
liujed: Good point. Deletion of a service is not something I've thought about.

daemon/internal/cloud_client/change_active_traces_event.go
liujed: I'm explicitly copying it to avoid races. The safety we get seems worth the cost.

daemon/internal/cloud_client/change_active_traces_event.go
liujed: I've put most of the logging in Debug because I'm not sure how useful these messages would be to our users. Would they understand what this meant if they saw this as a warning? What could they do to address the warning?

daemon/internal/cloud_client/trace_event_request.go
liujed: I'm using the braces to demarcate the region of code to which the `Send the result to the client` comment applies.

daemon/internal/cloud_client/change_active_traces_event.go
mgritter: My goal is that if the user has a problem, we have some chance of diagnosing it from the existing logs alone. This means putting errors in Warning or Info, even if we don't expect the user to be able to do anything about it, simply to avoid "can you turn debug on and make it happen again."  The only reason not to run in Debug mode all the time is space constraints. ​

A secondary goal is that when there *is* a problem, the user has some chance of noticing and contacting us about it, which they can't do if they're silenced by default. While it would be better to have end-to-end telemetry that either we or the user could monitor, "hey, my log is getting spammed by X" does happen.

Because we don't actually have an alerting mechanism (nobody is getting paged or emailed when things go really wrong), I think we should err on the side of putting more into the log as it's low-cost and serves both these goals.

daemon/internal/cloud_client/change_active_traces_event.go
liujed: Addressed in da12c1b.

daemon/internal/cloud_client/cloud_client.go
liujed: Feasible, sure. But where would it go? I'd like to avoid having an externally function visible that shouldn't be called anywhere else.

daemon/internal/cloud_client/long_poll_service_event.go
liujed: We don't always want to resume long polling—the handler for `changeActiveTracesEvent` sometimes returns without re-enqueuing the long poll. I'll add comments to clarify where the responsibility lies.

daemon/internal/cloud_client/cloud_client.go
mgritter: The place I'm coming from is that inevitably somebody wants to write a unit test and the inline function is too hard to get to with the right arguments or mocking.  (I've been that person, with this example: https://github.com/hashicorp/vault/blob/2a8c36be280da6d0a1303c814a8d05ce274e5414/vault/acl.go#L523)

It can be package-scoped (not exported) so it's not entirely "visible".

Breaking it out also makes it easier to interpret backtraces, particularly if there are multiple funcs() involved.

daemon/internal/cloud_client/long_poll_service_event.go
liujed: Done in 1190f53.

daemon/internal/cloud_client/client_state.go
liujed: Done in 1190f53.

daemon/internal/cloud_client/long_poll_service_event.go
liujed: Done in 9eca8de.

daemon/internal/cloud_client/cloud_client.go
liujed: Done in 9eca8de.

daemon/internal/cloud_client/cloud_client.go
liujed: Done in 9eca8de.

daemon/internal/cloud_client/cloud_client.go
liujed: Done in 9eca8de.

daemon/internal/cloud_client/cloud_client.go
liujed: Done in 9eca8de.

daemon/internal/cloud_client/change_active_traces_event.go
liujed: Done in 9eca8de.

mgritter (approved):

daemon/internal/cloud_client/cloud_client.go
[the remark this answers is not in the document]
liujed: Let's defer handling of service deletion.

PR #22, opened 2021-05-01
Tweaks to daemon–middleware API.

liujed:
Bodies of registration requests from middleware now contain just a single JSON object. Previously, it was a stream of objects, which some frameworks (e.g., Express.js) don't appear to support very well.

When sending trace events, middleware now indicate whether to expect more events in future requests.

Wait for middleware to finish sending their events before shutting down a trace.

Instrumented backend collector with a packet count summary. We now report the number of requests and responses uploaded when shutting down a trace.

Added a missing channel flush.

Added more logging.


mgritter (approved):
Looks OK.

PR #24, opened 2021-05-06
Ensure periodicFlush exits

mgritter:
"break" only exits the "select" statement, not the loop.

The loop ran forever busy-waiting because flushDone was always triggered once closed.

### DIFF
--- a/trace/backend_collector_test.go
+++ b/trace/backend_collector_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/akitasoftware/akita-libs/akid"
 	"github.com/akitasoftware/akita-libs/akinet"
 	kgxapi "github.com/akitasoftware/akita-libs/api_schema"
+	"github.com/akitasoftware/akita-libs/batcher"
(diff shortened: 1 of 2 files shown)


trace/backend_collector_test.go
```diff
@@ -89,7 +90,7 @@ func TestObfuscate(t *testing.T) {
 		},
 	}
 
-	col := NewBackendCollector(fakeSvc, fakeLrn, mockClient, kgxapi.Inbound, nil)
+	col := NewBackendCollector(fakeSvc, fakeLrn, mockClient, kgxapi.Inbound, nil, nil)
```
liujed, line 93: How did this compile before?

liujed (approved):
Thanks for the quick fix!

trace/backend_collector_test.go
mgritter: It didn't!  We need to enable CircleCI test runs on the akita-cli repository to catch things like this.

PR #25, opened 2021-05-06
Add a diagram showing daemon network traffic

mgritter:
Seems hard to view the picture on the PR?  Here it is:

Daemon interactions


liujed:
This works for me for viewing the PNG:
image

liujed (approved):
Looks good. A couple nits:
* The `{service}`s are different on each side of the daemon. On the middleware side, they are service names, whereas on the cloud side, they are service IDs. Should clarify this; maybe use `{serviceID}` for the cloud side.
* After the middleware closes off its event stream, the daemon flushes its backend collector, so there might be another trace-event message to the cloud.

PR #33, opened 2021-05-21
Convert strings in traces to other types depending on context

liujed:
When producing IR from traces, don't auto-convert strings to ints and bools. This was causing us to generate incorrect specs.


thatplguy (approved):

liujed:
We need to selectively enable the auto-conversion. Seems like it's desired in some contexts (e.g., query-encoded inputs), but not in others (e.g., JSON bodies).

mgritter (approved):

PR #40, opened 2021-06-07
Tag traces and models with their source

liujed:
Traces and models are now tagged to indicate whether they were obtained from CI, were manually created by a user, or were uploaded.

Also introduced a `tags` package that acts as a central repository for all our tags.

To do: figure out how to tag traces and models that were obtained from a staging/production deployment. Currently, deployment traces are created manually before they are populated with events by the daemon. Similarly, deployment models are also created manually from deployment traces. For now, we can tag these with command-line arguments when they are created.


thatplguy (approved):

mgritter (approved):
Looks good.

thatplguy (approved):

PR #42, opened 2021-06-09
Rate limiting for apidump and learn

mgritter:
This change provides a --rate-limit CLI argument that sets the maximum throughput in witnesses/minute.  Samples are collected during a random interval within each 5-minute epoch, so that we get a continguous sample rather than a uniform sample across all requests.


trace/rate_limit.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	printer.Debugf("Expired %v old requests\n", expired)
+}
+
+func NewRateLimit(witnessesPerMinute float64) *SharedRateLimit {
+	witnessLimit := witnessesPerMinute * viper.GetDuration(RateLimitEpochTime).Minutes()
+	if witnessLimit < 1 {
+		printer.Warningln("Witnesses per minute rate is too low, rounding up to 1 per 5 minutes.")
+		witnessLimit = 1
+	}
```
liujed, line 281: Doesn't this round up to one witness per `viper.GetDuration(RateLimitEpochTime)`?

trace/rate_limit.go
```diff
@@ -0,0 +1,294 @@
...
+			expired += 1
+		}
+	}
+	printer.Debugf("Expired %v old requests\n", expired)
+}
+
+func NewRateLimit(witnessesPerMinute float64) *SharedRateLimit {
+	witnessLimit := witnessesPerMinute * viper.GetDuration(RateLimitEpochTime).Minutes()
+	if witnessLimit < 1 {
+		printer.Warningln("Witnesses per minute rate is too low, rounding up to 1 per 5 minutes.")
```
liujed, line 279: (Nit: comma splice)

trace/rate_limit.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	r := &SharedRateLimit{
+		WitnessesPerMinute: witnessesPerMinute,
+		WitnessesPerEpoch:  int(witnessLimit),
+		FirstEstimate:      true,
+		done:               make(chan struct{}),
+	}
+
+	// Start the first epoch.
+	// run() will start an interval immediately.
```
liujed, line 290: `run()` should have a doc comment establishing this as part of its contract.

trace/rate_limit.go
```diff
@@ -0,0 +1,294 @@
...
+	// Trigger request expiration for all collectors
+	threshold := epochStart.Add(-1 * viper.GetDuration(RateLimitMaxDuration))
+	for _, child := range r.children {
+		child.epochCh <- threshold
+	}
+}
+
+// Check if request should be sampled; increase the count by one.
+func (r *SharedRateLimit) AllowHTTPRequest() bool {
+	r.lock.Lock()
```
liujed, line 155: Are we concerned that this might become a bottleneck? (If yes, alas, I don't see an obvious design that avoids this.)

liujed (approved):

trace/rate_limit.go
mgritter: Somewhat, but I suspect in practice I expect most traffic comes through a single interface so there is not a lot of contention.

I initially thought of doing something more complicated using channels. You can give every collector a channel of size 1 (or, larger than 1), which it must successfully read from before it can record a packet.  It does the read in a nonblocking fashion and drops the packet if it fails.  The share rate limit farms out its budget to any channel that is writeable.  This means the control is not as precise, but that seems acceptable.

The problem with implementing this in Go is that there's no good way to select() on an arbitrary list of channels, as far as I know.  You could do a fan-out using multiple goroutines reading from a single source channel.

A better way is probably batching-- grab N "leases" to send a packet, when the lock is grabbed, instead of just 1, and decrement the count locally per-collector.  If needed, these leases could have an expiration time after which they need to be returned to the pool.

liujed: Yeah, these options are sufficiently complicated that I think we can just keep them in our back pockets for if/when the bottleneck becomes an issue.

trace/rate_limit.go
mgritter: Yes, that's correct-- I didn't think it was worth trying to nicely format the configured value (or introduce the term "epoch".)

trace/rate_limit.go
mgritter: fixed.

trace/rate_limit.go
mgritter: Fixed

trace/rate_limit.go
```diff
@@ -0,0 +1,294 @@
...
+	r.SampleIntervalActive = true
+}
+
+func (r *SharedRateLimit) IntervalStarted() bool {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	return r.SampleIntervalActive
+}
+
+func (r *SharedRateLimit) endInterval(end time.Time) {
```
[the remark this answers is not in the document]
mgritter: Fixed.

thatplguy (approved):
LGTM 👍

PR #47, opened 2021-06-16
Only generate ClientID once per CLI instance

liujed:


cmd/internal/akiflag/root.go
```diff
@@ -1,7 +1,12 @@
...
+import "github.com/akitasoftware/akita-libs/akid"
+
 // A set of variables holding the values of global flags exposed by the root
 // command. This allows us to share those values with subcommands without
 // creating an import loop.
 
 var Domain string
+
+// A global ClientID, generated once per CLI instance.
+var ClientID akid.ClientID = akid.GenerateClientID()
```
mgritter, line 12: I am get nervous when I see this, only because I don't know the rules for Go on when things get initialized.

So I looked it up; for package-level variables it's in declaration order within each file, and lexicographic order of files within a package.  Packages are initialized in dependency order.  So I think we are OK here, but I would feel a tiny bit better if this were a function using local state, rather than a package-level variable.

mgritter (approved):
Looks like this is incorporated in the other PR already.

cmd/internal/akiflag/root.go
liujed: Done in 87fab38.

PR #48, opened 2021-06-16
Guard commands when running in CI

liujed:
When running in CI as part of a GitHub PR, check whether the PR is Akita-enabled before doing anything. If the PR is not Akita-enabled, print a warning and exit (with code 0, to avoid killing any CI scripts) without further interacting with Akita Cloud.


cmd/root.go
```diff
@@ -123,16 +124,17 @@ func init() {
 		viper.BindPFlag("verbose-level", flag.CommandLine.Lookup("v"))
 	}
 
-	// Register subcommands.
-	rootCmd.AddCommand(apidiff.Cmd)
-	rootCmd.AddCommand(apidump.Cmd)
-	rootCmd.AddCommand(apispec.Cmd)
+	// Register subcommands. Most commands that interact with Akita Cloud should
+	// be guarded by ci_guard.
+	rootCmd.AddCommand(ci_guard.GuardCommand(apidiff.Cmd))
```
mgritter, line 129: I'm not sure why apidiff is on the list?  (I'm not sure it would be used in a CI run anwyay, so it doesn't matter.)

mgritter (approved):

cmd/root.go
liujed: It interacts with Akita Cloud and consumes cloud resources to produce the diff. It's also not a purely interactive command: you can write the diff to a file with `--out`.

PR #49, opened 2021-06-17
Convert environment variables to Kubernetes deployment tags

mgritter:
I discovered while testing that when using 'learn' that the trace did not have the full set of tags; only the spec did.  I fixed this for deployment; should it also be fixed for CI?

Probably, I think, the automatic tags creation should be done by the original command during command line parsing; having "learn" process the tags twice, once in apidump.Run() and once in apispec.Run() and once in  is a little off.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
+	"github.com/akitasoftware/akita-cli/deployment"
 	"github.com/akitasoftware/akita-cli/location"
 	"github.com/akitasoftware/akita-cli/plugin"
 	"github.com/akitasoftware/akita-cli/printer"
@@ -173,6 +174,23 @@ func Run(args Args) error {
 	}
 	if args.Filter != "" {
 		traceTags[tags.XAkitaDumpFilterFlag] = args.Filter
+	}
+
+	// Import information about production or staging environment
+	// FIXME: this is performed in apispec as well; does that cause problems if the same
+	// args.Tags map is re-used?
+	{
+		deploymentType, deploymentTags := deployment.GetDeploymentInfo()
+
+		// Override user type, but not CI.
+		if deploymentType != deployment.None &&
+			args.Tags[tags.XAkitaSource] == tags.UserSource || args.Tags[tags.XAkitaSource] == tags.Source("") {
+			args.Tags[tags.XAkitaSource] = tags.DeploymentSource
+		}
+
+		for k, v := range deploymentTags {
+			args.Tags[k] = v
+		}
 	}
 
 	// Build path filters.
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
+	"github.com/akitasoftware/akita-cli/ci"
 	"github.com/akitasoftware/akita-cli/deployment"
 	"github.com/akitasoftware/akita-cli/location"
 	"github.com/akitasoftware/akita-cli/plugin"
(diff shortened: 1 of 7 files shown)


deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/akitasoftware/akita-cli/printer"
+	"github.com/akitasoftware/akita-libs/tags"
+)
+
+type Deployment string
+
+const (
+	None       Deployment = ""
+	Unknown               = "unknown"
+	Kubernetes            = "kubernetes"
```
liujed, line 15: I think Go wants us to specify `Deployment` on all declarations here; otherwise, only the first one is typed as `Deployment`, and the others are just untyped strings. Alas, the typing magic of `iota` doesn't apply here.

deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+		tagset[tags.XAkitaGitCommit] = c
+	}
+
+	if GetAWSTags(tagset) {
+		printer.Infof("Found AWS environment variables.\n")
+	}
+
+	if GetKubernetesTags(tagset) {
+		printer.Infof("Found Kubernetes environment variables.\n")
+		deploymentType = Kubernetes
```
liujed, line 46: This overwrites the deployment type obtained from `AKITA_DEPLOYMENT` in the environment. Is this intentional?

deployment/deployment.go
```diff
@@ -0,0 +1,85 @@
...
+		EnvVar string
+		Key    tags.Key
+	}{
+		{"AKITA_K8S_NAMESPACE", tags.XAkitaKubernetesNamespace},
+		{"AKITA_K8S_NODE", tags.XAkitaKubernetesNode},
+		{"AKITA_K8S_HOST_IP", tags.XAkitaKubernetesHostIP},
+		{"AKITA_K8S_POD", tags.XAkitaKubernetesPod},
+		{"AKITA_K8S_POD_IP", tags.XAkitaKubernetesPodIP},
+		{"AKITA_K8S_DAEMONSET", tags.XAkitaKubernetesDaemonset},
+	}
```
liujed, line 64: Would be nice if we could have the environment–tag mapping in one place. Right now, it's spread across three functions: here, in `GetDeploymentInfo`, and in `GetAWSTags`.

What if we had a two-level map at the package level? Something like:
```
type mappingGroup int
const (
	defaultGroup = iota
	awsGroup
	kubernetesGroup
)

// Each value maps a tag key to its corresponding environment variable.
var envMapping = map[mappingGroup]map[tags.Key]string{
	defaultGroup: {
		tags.XAkitaDeployment: "AKITA_DEPLOYMENT",
		tags.XAkitaGitCommit: "AKITA_DEPLOYMENT_COMMIT",
	},
	awsGroup: {
		tags.XAkitaAWSRegion: "AKITA_AWS_REGION",
	},
	kubernetesGroup: {
		tags.XAkitaKubernetesNamespace: "AKITA_K8S_NAMESPACE",
		...
	},
}
```
Then, the job of extracting tags from the environment can be factored into something that looks like `GetKubernetesTags`, but also called from the other two places.

apidump/apidump.go
```diff
@@ -175,6 +176,23 @@ func Run(args Args) error {
...
 
+	// Import information about production or staging environment
+	// FIXME: this is performed in apispec as well; does that cause problems if the same
+	// args.Tags map is re-used?
+	{
+		deploymentType, deploymentTags := deployment.GetDeploymentInfo()
+
+		// Override user type, but not CI.
+		if deploymentType != deployment.None &&
+			args.Tags[tags.XAkitaSource] == tags.UserSource || args.Tags[tags.XAkitaSource] == tags.Source("") {
```
liujed, line 187: I think you want parens around the `||` sub-expression: `&&` has higher precedence.

apispec/run.go
```diff
@@ -99,6 +100,21 @@ func Run(args Args) error {
 		}
 	}
 
+	// Import information about production or staging environment
+	{
+		deploymentType, deploymentTags := deployment.GetDeploymentInfo()
+
+		// Override user type, but not CI.
+		if deploymentType != deployment.None &&
+			args.Tags[tags.XAkitaSource] == tags.UserSource {
```
liujed, line 109: This looks different from what is in `apidump.go`. Is this intentional? If not, maybe we should factor this into its own method somewhere.

mgritter: It is intentional, because here Tags[tags.XAkitaSource] has been explicitly initialized to UserSource above.

But, I am not too happy with the result.  I think the right thin in apispec is:

* if CI and source not set, set to CI
* if production and source not set, set to production
* if source not set, set to user

Instead of trying to do it in the reverse order.

deployment/deployment.go
mgritter: Hm, I'm not sure why it compiled, but you're correct.

deployment/deployment.go
mgritter: I wish I had come up with a synonym. Any suggestions?

The user can't set the deploymentType, which is only used internally; it doesn't even appear as a tag (though maybe the lesson is that it should!) I figured the presence of Kubernetes-specific tags was sufficient.

They can set the x-akita-deployment value, which will be "staging" or "production" or "test1" etc. to distinguish different clusters.

apidump/apidump.go
mgritter: Yup, thanks.

deployment/deployment.go
mgritter: Good idea, implemented.

apispec/run.go
mgritter: Rewrote this entirely.

deployment/deployment.go
liujed: Looks like this is no longer an issue with this rewrite.

liujed (approved):
Looks good!

deployment/deployment.go
```diff
@@ -0,0 +1,100 @@
...
+const (
+	None       Deployment = ""
+	Any        Deployment = "any"
+	Unknown    Deployment = "unknown"
+	AWS        Deployment = "aws"
+	Kubernetes Deployment = "kubernetes"
+)
+
+// Map of environment variables to tags, grouped by deployment type.
+var environmentToTag map[Deployment]map[string]tags.Key = map[Deployment]map[string]tags.Key{
```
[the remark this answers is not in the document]
mgritter: Hm... I'll leave it be for now, my expectation is that the opposite is likely to be true (we accept more than one environment variable name.)  Obviously we could just make it a list of pairs too.

PR #54, opened 2021-06-23
Add --allow-host and --allow-path flags

mgritter:
Only add filters that are enabled to the stack of collectors.
Add a pre-filter packet count so that we can compare and warn if all
packets we eliminated due to filters.

```
[INFO] Captured 8 HTTP requests before allow and exclude rules, but all were filtered.
[ERROR] No inbound HTTP calls captured! 🛑
```

### DIFF
--- a/trace/filters.go
+++ b/trace/filters.go
@@ -10,6 +10,7 @@ import (
 )
 
 // Filters out HTTP paths.
+// TODO: compile the N regular expressions into one for efficiency.
 func NewHTTPPathFilterCollector(matchers []*regexp.Regexp, col Collector) Collector {
 	return &genericRequestFilter{
 		Collector: col,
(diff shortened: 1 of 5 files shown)


mgritter:
I was wondering if we should add the allow/exclude filters to some tags on the spec, like we do for the network filter argument.

https://github.com/akitasoftware/akita-cli/blob/0896319d67742cfd6e5966d254b79d084e2f1bdb/apidump/apidump.go#L156-L163

trace/filters.go
```diff
@@ -40,6 +42,39 @@ func NewHTTPHostFilterCollector(matchers []*regexp.Regexp, col Collector) Collec
...
+		filterFunc: func(r akinet.HTTPRequest) bool {
+			for _, m := range matchers {
+				if m.MatchString(r.Host) {
+					return true
+				}
+			}
+			return false
+		},
+	}
+}
```
liujed, line 76: Seems a shame that we can't just build the corresponding deny filter and invert it. Another time, perhaps.

liujed (approved):
Looks good!

liujed:
> I was wondering if we should add the allow/exclude filters to some tags on the spec, like we do for the network filter argument.
> 
> https://github.com/akitasoftware/akita-cli/blob/0896319d67742cfd6e5966d254b79d084e2f1bdb/apidump/apidump.go#L156-L163

Would be nice. I don't know whether these tags are used, though.

PR #59, opened 2021-06-29
Capture all traffic if --dogfood is set.

mgritter:
Instead of adding a header that allows capture, disable the filter when "--dogfood" is set.

This is not sufficient by itself, a back-end change needs to eliminate the filter too: https://github.com/akitasoftware/superstar/pull/348


mgritter:
I tried removing the dependency to spec_util entirely, but as the code stands now we need two versions of ContainsCLITraffic, one that works on ParsedNetworkTraffic and one that works on witnesses. :(

Still, it feels like our special headers should live somewhere besides spec_util, but at least all the references are to the same place (I think.)

learn/learn_runner.go
```diff
@@ -120,8 +107,10 @@ func createHARFile(in <-chan akinet.ParsedNetworkTraffic, interfaceName string,
 
 	l := har.NewLogger()
 	for t := range in {
-		if containsCLITraffic(t.Content) {
-			continue
+		if util.ContainsCLITraffic(t) {
+			if !viper.GetBool("dogfood") {
```
liujed, line 111: Flip these conditions? Checking the boolean seems cheaper than checking `ContainsCLITraffic`.

liujed (approved):

learn/learn_runner.go
mgritter: Probably, but for non-Akita users the situation is reversed: the GetBool is strictly overhead, so it should be executed as few times as possible.

liujed: Wouldn't branch prediction take care of that overhead? But, yes, we shouldn't be optimizing for dogfooding.

PR #62, opened 2021-07-02
Increase upload batch size to 120 witnesses

liujed:
This is just a stopgap. We really should be measuring the size of the accumulated payload, not the number of witnesses.

### DIFF
--- a/trace/backend_collector.go
+++ b/trace/backend_collector.go
@@ -31,7 +31,7 @@ const (
 	pairCacheCleanupInterval = 30 * time.Second
 
 	// Max size per upload batch.
-	uploadBatchMaxSize = 10
+	uploadBatchMaxSize = 120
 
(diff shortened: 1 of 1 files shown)


mgritter (approved):
I wish the batching were done over all collectors, not just one, because I worry about total memory usage.  But, as long as gopacket is giving us a 1000-packet buffer, this will not be the tall pole in the tent.

120 request/response pairs @ 4KB average size (unserialized?) = 480KB

Times 10 (?) max host interfaces = 4.8MB.

Completely reasonable unless my numbers are way off.  Once we do some profiling we should have a better handle on it.

liujed:
The issue with counting witnesses is we get fragmentation when the batch gets put onto SQS. Currently, `front` treats each request in isolation, so it doesn't batch witnesses across requests.

PR #63, opened 2021-07-07
Add time-bounds options to `apispec`

liujed:
This adds `--from-time` and `--to-time` options to `apispec` for filtering which witnesses are included in the generated spec. The intent is for this to filter by the witnesses' observation time. However due to limitations in the current database schema, for witnesses that are observed multiple times, this currently filters on the first time each witness was observed in a learn session. Future back-end improvements will bring these options closer to the intended semantics.


thatplguy (approved):
LGTM.

mgritter (approved):
Looks good.

PR #67, opened 2021-07-13
Add hidden flag (and config file entry) for max HTTP length

mgritter:
Depends on https://github.com/akitasoftware/akita-libs/pull/58

Also lowers the stream timeout to 10 seconds, with a flag for configuring it.


liujed (approved):
LGTM

PR #69, opened 2021-07-14
Limit number of pages gopacket can allocate.

mgritter:


cmd/root.go
```diff
@@ -143,9 +143,14 @@ func init() {
 	rootCmd.PersistentFlags().MarkHidden("stream-timeout-seconds")
 	viper.BindPFlag("stream-timeout-seconds", rootCmd.PersistentFlags().Lookup("stream-timeout-seconds"))
 
+	// For explanation of these defaults see net_parse.go
+	rootCmd.PersistentFlags().IntVar(&pcap.MaxBufferedPagesTotal, "gopacket-pages", 150_000, "Maximum number of TCP reassembly pages to allocate")
+	rootCmd.PersistentFlags().MarkHidden("gopacket-pages")
+	rootCmd.PersistentFlags().IntVar(&pcap.MaxBufferedPagesPerConnection, "gopacket-per-conn", 4_000, "Maximum number of TCP reassembly pages per connectio")
```
liujed, line 149: Typo: "connectio"

liujed (approved):
LGTM

pcap/net_parse.go
```diff
@@ -18,9 +18,27 @@ import (
 // the data even if there is a gap in the collected sequence.
 var StreamTimeoutSeconds int64 = 10
 
+// Maximum size of gopacket reassembly buffers.
+//
+// A gopacket page is 1900 bytes.
+// We want to cap the total memory usage at about 300MB = 157894 pages
+var MaxBufferedPagesTotal int = 150_000
```
mgritter, line 25: Note that this would result in 300MB *per assembler* maximum, so it might still be too big as we create one per interface, and I think incoming vs. outcoming is handled is separate goroutines too.

On our production system each node has six interfaces, so 1800MB maximum if we have no filter argument, or 3600MB maximum if we have a filter argument.

There does not seem to be an easy way to apply a shared limit.

thatplguy (approved):

PR #70, opened 2021-07-15
Use case-insensitive lookups for service names

liujed:


mgritter (approved):
Looks good.

I had thought we had an endpoint in front that did name-to-id conversion, but it looks like that is only for trace names and spec names.

thatplguy (approved):

PR #71, opened 2021-07-15
Fix success message for akita upload cmd

thatplguy:
The CLI output for `akita upload` mixed `info` output on `stderr` and `stdout`, which occasionally caused the formatting to break.  This PR switches to use `stderr` exclusively.

### DIFF
--- a/upload/run.go
+++ b/upload/run.go
@@ -70,8 +70,6 @@ func Run(args Args) error {
 		ObjectName:  objectName,
 	}
 	printer.Stderr.Infof("%s 🎉\n", aurora.Green("Success!"))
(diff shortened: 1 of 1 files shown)


liujed (commented):
Does anyone depend on the old behaviour? Looks like this has been like this since `upload` was introduced in #4. (Of course, we should really just provide flags for switching to machine-readable output.)

liujed (approved):

thatplguy:
AFAIK no one depends on this.  I agree that the right thing is to provide flags for machine-readable output.  I'll look at that in a separate PR.

PR #72, opened 2021-07-16
Assign versions on model creation

thatplguy:
This PR adds a `--versions` flag to the `apispec` and `learn` commands, which
allows users to assign versions to models at creation time.

Example usage:

* `akita apispec --traces akita://akibox:trace:t1 --versions stable`
* `akita learn --service akibox --versions stable,demo`

### DIFF
--- a/apispec/run.go
+++ b/apispec/run.go
@@ -53,6 +53,7 @@ type Args struct {
 	Service                    string
 	Format                     string
 	Tags                       map[tags.Key]string
+	Versions                   []string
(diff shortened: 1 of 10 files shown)


mgritter (approved):
Looks OK to me.

Now that we have a reserved list, do we want to check it in the "setversion" command, too?
https://github.com/akitasoftware/akita-cli/blob/main/setversion/run.go

thatplguy:
Good idea, I'll add that.

PR #78, opened 2021-07-30
Upgrade gopacket to v1.1.19 (+ Akita changes)

mgritter:
The is the CLI version of https://github.com/akitasoftware/akita-libs/pull/65

I think we also need to update akita-libs in this repo to point at the one that has been upgraded, once that change has landed.


mgritter:
An update of akita-libs is not necessary. https://golang.org/ref/mod#minimal-version-selection

thatplguy (approved):

PR #79, opened 2021-08-03
Split the maximum page cache size up among each parser.

mgritter:
Split the configured page cache size up among the different parsers, one per interface and per direction.

This successfully reduces the memory; we don't have a way to measure what it does to "goodput" or if we lose some packets because we could not buffer enough to see the retransmission.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -388,10 +388,14 @@ func Run(args Args) error {
 				}
 			}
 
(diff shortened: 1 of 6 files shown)


mgritter:
Whoops, I forgot I was going to update the unit tests today before making this a PR.  :(

thatplguy (approved):

PR #81, opened 2021-08-03
Bump akita-libs to pick up memory efficiency improvements

thatplguy:


mgritter (approved):
Yup, that's the right changeset. :)

PR #84, opened 2021-08-06
Update 'learn-sessions checkpoint' to return URI rather than ID

thatplguy:
As part of this change, `learn-sessions checkpoint` now creates an API model with a randomly-generated name rather than `""`, matching the behavior of `apispec` when no name is specified.

### DIFF
--- a/cmd/internal/legacy/util.go
+++ b/cmd/internal/legacy/util.go
@@ -7,18 +7,21 @@ import (
 	"path"
 	"time"
 
+	"github.com/akitasoftware/akita-libs/akiuri"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
 	"github.com/akitasoftware/akita-cli/cmd/internal/akiflag"
 	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/rest"
+	"github.com/akitasoftware/akita-cli/util"
+
 	"github.com/akitasoftware/akita-libs/akid"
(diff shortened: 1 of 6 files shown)


cmd/internal/legacy/util.go
```diff
@@ -81,12 +94,13 @@ func startLearnSession(c rest.LearnClient, baseSpecRef *kgxapi.APISpecReference,
 	return lrn, nil
 }
 
-func checkpointLearnSession(c rest.LearnClient, lrn akid.LearnSessionID, timeout time.Duration) (akid.APISpecID, error) {
+func checkpointLearnSession(c rest.LearnClient, lrn akid.LearnSessionID, timeout time.Duration) (akid.APISpecID, string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
-	specID, err := c.CheckpointLearnSession(ctx, lrn)
+	name := util.RandomAPIModelName()
+	specID, err := c.CreateSpec(ctx, name, []akid.LearnSessionID{lrn}, rest.CreateSpecOptions{})
```
mgritter, line 101: Hooray!

mgritter (approved):
LGTM.

cmd/internal/legacy/util.go
liujed: KGX auto-generates a name when one isn't provided: https://github.com/akitasoftware/superstar/blob/2cb7c7b1239ea94a4a7ab8cde699a85c38014e2d/services/kings_cross/specs_db/pg.go#L90-L92. The response doesn't include this name, however.

PR #88, opened 2021-08-19
Handle any content type that has +json at the end as JSON.

mgritter:
I thought about adding more text/* types, but that quickly got complex as we already handle text/yaml and text/x-yaml so there would have to be exception lists.

I don't know if there are any application/* types that are actually JSON but don't indicate it; we could add them later.

Tested locally with akibox:

```
@app.get("/users/{user_id}/test")
async def get_user(user_id: str):
    if user_id not in users:
        raise HTTPException(status_code=404, detail="User not found")
    return Response(content='{"user":"test", "value":17}',
                    media_type="application/vnd.api+json",
                    status_code=200)
```

### DIFF
--- a/learn/parse_http.go
+++ b/learn/parse_http.go
@@ -295,8 +295,6 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
 	switch {
 	case strings.HasSuffix(mediaType, "+json"):
 		mediaType = "application/json"
-	case mediaType == "text/csv":
-		mediaType = "text/plain"
 	}
 
 	var bodyData *pb.Data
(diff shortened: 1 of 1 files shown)


mgritter:
Our OpenAPI3 spec view still says:

```
Response Schema: application/json
```

I think we do not record the original content-type anywhere in the IR? Might be good to tackle, eventually.

learn/parse_http.go
```diff
@@ -288,6 +288,17 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
...
 
+	// Rewrite media type to JSON for types encoded as JSON
+	// TODO: XML parsing
+	// TODO: application/json-seq (RFC 7466)?
+	// TODO: more text/* types
+	switch {
+	case strings.HasSuffix(mediaType, "+json"):
+		mediaType = "application/json"
+	case mediaType == "text/csv":
+		mediaType = "text/plain"
```
liujed, line 299: Why not add a follow-through case for `text/csv` in the `switch` below?

liujed (approved):
Small nit.

learn/parse_http.go
mgritter: Yeah... that would probably be better for this particular case, I originally tried to handle

strings.HasPrefix(mediaType, "text/")

and then edited it just down to this.

PR #92, opened 2021-09-01
Check the number of witnesses reported by the server

mgritter:
before continuing with `apispec` on a local HAR file.


thatplguy (approved):
LGTM

PR #106, opened 2021-11-08
Disable collection of outbound witnesses by default

liujed:
Anyone wishing to restore the old behaviour can use the hidden flag `--enable-outbound-collection`. When this flag is used, the user is warned that the flag might be removed in a future release.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -246,10 +246,15 @@ func compileRegexps(filters []string, name string) ([]*regexp.Regexp, error) {
 func Run(args Args) error {
 	args.lint()
 
+	// Capture outbound packets during debugging so we can report statistics for
+	// packets not matching the user's filters.
 	debugEnabled := viper.GetBool("debug")
+	capturingOutbound := args.EnableOutbound || debugEnabled
 
 	if args.EnableOutbound {
 		printer.Warningln("Enabling collection of outbound traffic. The \"--enable-outbound-collection\" flag is deprecated and may be removed in a future release.")
+	} else if debugEnabled {
+		printer.Debugln("Capturing filtered traffic for debugging.")
 	}
 
 	// Get the interfaces to listen on.
@@ -259,12 +264,12 @@ func Run(args Args) error {
 	}
 
 	// Build inbound and outbound filters for each interface.
-	inboundFilters, outboundFilters, err := createBPFFilters(interfaces, args.Filter, args.EnableOutbound || debugEnabled, 0)
+	inboundFilters, outboundFilters, err := createBPFFilters(interfaces, args.Filter, capturingOutbound, 0)
 	if err != nil {
 		return err
 	}
 	printer.Debugln("Inbound BPF filters:", inboundFilters)
-	if args.EnableOutbound {
+	if capturingOutbound {
 		printer.Debugln("Outbound BPF filters:", outboundFilters)
 	}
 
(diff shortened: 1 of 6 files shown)


apidump/apidump.go
```diff
@@ -245,19 +246,27 @@ func compileRegexps(filters []string, name string) ([]*regexp.Regexp, error) {
 func Run(args Args) error {
 	args.lint()
 
+	debugEnabled := viper.GetBool("debug")
+
+	if args.EnableOutbound {
+		printer.Warningln("Enabling collection of outbound traffic. The \"--enable-outbound-collection\" flag is deprecated and may be removed in a future release.")
+	}
```
mgritter, line 253: Can you add a debug statement here stating that we're opening extra packet captures to inspect filtered traffic?

I was confused at the interaction of debugEnabled and EnableOutbound, so it might be worth including the extra information in the output too.

apidump/apidump.go
```diff
@@ -245,19 +246,27 @@ func compileRegexps(filters []string, name string) ([]*regexp.Regexp, error) {
...
 
 	// Build inbound and outbound filters for each interface.
-	inboundFilters, outboundFilters, err := createBPFFilters(interfaces, args.Filter, 0)
+	inboundFilters, outboundFilters, err := createBPFFilters(interfaces, args.Filter, args.EnableOutbound || debugEnabled, 0)
 	if err != nil {
 		return err
 	}
 	printer.Debugln("Inbound BPF filters:", inboundFilters)
-	printer.Debugln("Outbound BPF filters:", outboundFilters)
+	if args.EnableOutbound {
```
mgritter, line 267: args.EnabledOutbound || debugEnabled?  Or if len(outboundFilters) > 0 to cover both those cases? Or maybe just make it unconditional?

apidump/apidump.go
```diff
@@ -457,10 +472,20 @@ func Run(args Args) error {
...
+	if unfiltered {
 		printer.Stderr.Warningf("%s\n", printer.Color.Yellow("--filter flag is not set, this means that:"))
 		printer.Stderr.Warningf("%s\n", printer.Color.Yellow("  - all network traffic is treated as your API traffic"))
-		printer.Stderr.Warningf("%s\n", printer.Color.Yellow("  - outbound witness collection is disabled"))
+		if args.EnableOutbound {
+			printer.Stderr.Warningf("%s\n", printer.Color.Yellow("  - outbound witness collection is disabled"))
+		}
 	}
 
 	var stopErr error
```
mgritter, line 491: I think we need to change this (below the window of your changes) based on EnableOutbound?  It works when `--debug` is enabled, but the check for 0 is inappropriate otherwise.

```
			if outboundSummary.Total().TCPPackets == 0 {
				printer.Stderr.Infof("Did not capture any TCP packets during the trace.\n")
				printer.Stderr.Infof("%s\n", printer.Color.Yellow("This may mean the traffic is on a different interface, or that"))
				printer.Stderr.Infof("%s\n", printer.Color.Yellow("there is a problem sending traffic to the API."))
			} else {
				printer.Stderr.Infof("Did not capture any TCP packets matching the filter.\n")
				printer.Stderr.Infof("%s\n", printer.Color.Yellow("This may mean your filter is incorrect, such as the wrong TCP port."))
			}
```

mgritter (approved):
Looks good, maybe fix up the diagnostic code to handle outbound count == 0 being expected now?

apidump/apidump.go
liujed: In 5e0a89a. I've also introduced `capturingOutbound := args.EnableOutbound || debugEnabled` with a comment at the top of the function to help clarify.

apidump/apidump.go
liujed: Changed to `if capturingOutbound` in 5e0a89a.

apidump/apidump.go
liujed: Fixed in 5e0a89a to also guard on `capturingOutbound`.

PR #109, opened 2021-11-09
Remove error messages when an assembly context is nil.

mgritter:
Includes a (disabled) unit test demonstrating one way this can arise.
Added counters reporting number of such errors at the end of 'apidump'.

Question: is there a nice counter library we should be using?  Or is everything StatsD-based in the modern era?

### DIFF
--- a/pcap/stream.go
+++ b/pcap/stream.go
@@ -3,6 +3,7 @@ package pcap
 import (
 	"encoding/binary"
 	"net"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/gopacket"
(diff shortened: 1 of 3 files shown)


pcap/stream.go
```diff
@@ -104,7 +119,17 @@ func (f *tcpFlow) reassembledWithIgnore(ignoreCount int, sg reassembly.ScatterGa
 			acForFirstByte := sg.AssemblerContext(ignoreCount + int(discardFront))
 			ctx, ok := acForFirstByte.(*assemblerCtxWithSeq)
 			if !ok {
-				printer.Errorf("received AssemblerContext without TCP seq info, treating %s data as raw bytes\n", fact.Name())
+				// Previous we errored in this case:
```
liujed, line 122: typo: "Previously"

liujed (approved):

PR #111, opened 2021-11-10
Handle HTTP and 'other' body types

mgritter:
Depends on: https://github.com/akitasoftware/akita-ir/pull/6 and https://github.com/akitasoftware/akita-libs/pull/96

I'm wondering if this will cause us to spend a lot of time examining things like png or jpg images returned from an endpoint.

One mitigation is to tune MaxBufferedBody way, way down, it's currently at 5MB.

I didn't feel comfortable making the body a "None" in a "Oneof None String" field.  This will at least give us some information about whether small response bodies are identical.  But I'd be happy to come up with a different way of signaling "we didn't expect to understand the body format, so we didn't try."

### DIFF
--- a/learn/parse_http.go
+++ b/learn/parse_http.go
@@ -358,19 +358,40 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
 			return nil, errors.Wrapf(err, "could not parse YAML body")
 		}
 		pbContentType = pb.HTTPBody_YAML
+	case "text/html":
+		body, err := limitedBufferBody(bodyStream)
+		if err != nil {
+			return nil, err
+		}
+		// This is a fairly cheesy way of handling the body, is there a better one?
+		bodyData = parseElem(string(body), spec_util.INTERPRET_STRINGS)
+		pbContentType = pb.HTTPBody_TEXT_HTML
 	case "multipart/form-data":
 		return parseMultipartBody("form-data", mediaParams["boundary"], bodyStream, statusCode)
 	case "multipart/mixed":
 		return parseMultipartBody("mixed", mediaParams["boundary"], bodyStream, statusCode)
 	default:
-		return nil, ParseAPISpecError(fmt.Sprintf("could not parse body with media type: %s", mediaType))
+		// If an unknown type, handle it like text/plain or text/html, but record
+		// the original content-type in the witness.
+		body, err := limitedBufferBody(bodyStream)
+		if err != nil {
+			return nil, err
+		}
+		bodyData = parseElem(string(body), spec_util.INTERPRET_STRINGS)
+		pbContentType = pb.HTTPBody_OTHER
+	}
+
+	bodyMeta := &pb.HTTPBody{
+		ContentType: pbContentType,
+	}
+
+	if pbContentType == pb.HTTPBody_OTHER {
+		bodyMeta.OtherType = mediaType
 	}
 
 	httpMeta := &pb.HTTPMeta{
 		Location: &pb.HTTPMeta_Body{
-			Body: &pb.HTTPBody{
-				ContentType: pbContentType,
-			},
+			Body: bodyMeta,
 		},
 		ResponseCode: int32(statusCode),
 	}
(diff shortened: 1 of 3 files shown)


learn/parse_http.go
```diff
@@ -358,19 +358,40 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
 			return nil, errors.Wrapf(err, "could not parse YAML body")
 		}
 		pbContentType = pb.HTTPBody_YAML
+	case "text/html":
+		body, err := limitedBufferBody(bodyStream)
+		if err != nil {
+			return nil, err
+		}
+		// This is a fairly cheesy way of handling the body, is there a better one?
+		bodyData = parseElem(string(body), spec_util.INTERPRET_STRINGS)
```
liujed, line 367: Seems not worth trying to interpret strings: if it claims to be HTML, it will probably have tags in it, and is therefore unlikely to be successfully parsed as an int, float, or bool.

learn/parse_http.go
```diff
@@ -358,19 +358,40 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
...
 		return parseMultipartBody("mixed", mediaParams["boundary"], bodyStream, statusCode)
 	default:
-		return nil, ParseAPISpecError(fmt.Sprintf("could not parse body with media type: %s", mediaType))
+		// If an unknown type, handle it like text/plain or text/html, but record
+		// the original content-type in the witness.
+		body, err := limitedBufferBody(bodyStream)
+		if err != nil {
+			return nil, err
+		}
+		bodyData = parseElem(string(body), spec_util.INTERPRET_STRINGS)
```
liujed, line 380: Handling like `application/octet-stream` seems more appropriate. Similarly to HTML, it doesn't seem worth trying to interpret strings here.

learn/parse_http.go
```diff
@@ -358,19 +358,40 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
...
+		pbContentType = pb.HTTPBody_OTHER
```
liujed, line 381: This could be unified with the `application/octet-stream`, `text/plain`, `text/csv`, and `text/html` cases with a nested switch (or a function call) to determine `pbContentType` and whether strings should be interpreted.

liujed (approved):
Looks good, though some nits regarding string interpretation and code duplication.

liujed:
> I'm wondering if this will cause us to spend a lot of time examining things like png or jpg images returned from an endpoint.

It looks like we already have this issue for `application/octet-stream`!

mgritter:
Some examples from local testing:

```
  responses {
    key: "Kibd55-BvCg="
    value {
      primitive {
        bytes_value {
          value: "O\307@\ry9\r\273J\322\330\032\020\344B\300\023\n\242N=n\362\360\206hS\021\242Y\324\331"
        }
      }
      meta {
        http {
          body {
            content_type: OTHER
            other_type: "image/vnd.microsoft.icon"
          }
          response_code: 200
        }
      }
    }
  }
  
  responses {
    key: "UdKRQCGhl9o="
    value {
      primitive {
        string_value {
          value: "i_khiXe40uflLyqmOln3fn2VQ629E-5RmOUjfvspyJI="
        }
      }
      meta {
        http {
          body {
            content_type: TEXT_HTML
          }
          response_code: 200
        }
      }
    }
  }
```

learn/parse_http.go
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/111/commits/b88090ba52de7c92877db4ea862c3fbbd69c7f1b

learn/parse_http.go
mgritter: Refactored in https://github.com/akitasoftware/akita-cli/pull/111/commits/b88090ba52de7c92877db4ea862c3fbbd69c7f1b

mgritter:
Re-requesting review after refactoring

mgritter:
D'oh!  octet-stream is actually _slightly_ different.

liujed (approved):

PR #113, opened 2021-11-13
Update libs

liujed:
Removed use of `api_schema.Outbound`. Replaced some uses of "inbound" and "outbound" with language about filters (e.g., "user filter" for the inbound filter and "negative filter" for the outbound filter, and "matched" vs. "unmatched").


liujed:
Depends on akitasoftware/akita-libs#99.

mgritter (approved):
Thank you!

PR #114, opened 2021-11-18
Initial TLS support

liujed:
Parse TLS handshakes and report them to the back end.

Augmented the service graph with TLS edges, detailing the TLS version and application-layer protocol, if known.

### DIFF
--- a/cmd/internal/get/graph.go
+++ b/cmd/internal/get/graph.go
@@ -184,6 +184,14 @@ func getGraph(cmd *cobra.Command, args []string) error {
 			}
 		}
 		resp.TCPEdges = replacementTCPEdges
+
+		replacementTLSEdges := make([]api_schema.TLSGraphEdge, 0, len(resp.TLSEdges))
+		for _, e := range resp.TLSEdges {
+			if e.Source != "" && e.Target != "" {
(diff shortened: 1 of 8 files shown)


liujed:
Depends on akitasoftware/akita-libs#100.

cmd/internal/get/graph.go
```diff
@@ -293,6 +342,26 @@ func printGraphByTarget(graph *api_schema.GraphResponse) {
...
 		return false
 	})
+	sort.Slice(graph.TLSEdges, func(i, j int) bool {
+		edgeI, edgeJ := graph.TLSEdges[i], graph.TLSEdges[j]
+		if edgeI.Source != edgeJ.Source {
+			return edgeI.Source < edgeJ.Source
+		}
+		if edgeI.Target != edgeJ.Target {
+			return edgeI.Target < edgeJ.Target
+		}
```
mgritter, line 352: Because this is printGraphByTarget, shouldn't target be the first key to sort on?

mgritter (approved):
Looks good.

cmd/internal/get/graph.go
liujed: Good catch. Fixed in aa1744b.

PR #118, opened 2022-01-31
Add tracking stats to witnesses

thatplguy:
Depends on https://github.com/akitasoftware/akita-ir/pull/8.


liujed:
Closing. The IR size increase turned out to be too much, and is not currently relevant.

PR #128, opened 2022-04-29
Update libs

liujed:
Also fixed a latent bug where timelines always had 0 latency

### DIFF
--- a/rest/learn_client.go
+++ b/rest/learn_client.go
@@ -202,14 +202,15 @@ func (c *learnClientImpl) SetSpecVersion(ctx context.Context, specID akid.APISpe
 	return c.post(ctx, path, req, &resp)
 }
 
-// Returns individual events
-func (c *learnClientImpl) GetUnaggregatedTimeline(ctx context.Context, serviceID akid.ServiceID, deployment string, start time.Time, end time.Time, limit int) (kgxapi.TimelineResponse, error) {
+// Returns events aggregated in 1-minute intervals.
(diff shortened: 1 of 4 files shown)


liujed:
Depends on akitasoftware/akita-libs#130.

rest/learn_client.go
```diff
@@ -214,6 +215,7 @@ func (c *learnClientImpl) GetUnaggregatedTimeline(ctx context.Context, serviceID
 	q.Add("key", "method")
 	q.Add("key", "path")
 	q.Add("key", "code")
+	q.Add("aggregate", string(api_schema.Aggr_99p))
```
thatplguy, line 218: Does the capability to get an unaggregated timeline still exist in the new version?  IIRC this meant retrieving one data point per witness, which is presumably why P99 was omitted.

In the new version, this probably returns events at the granularity of the atomic bucket size?

I think that's fine, and in that case it makes sense to add P99 latency.  For individual events, P99 latency wouldn't have made sense.

thatplguy (approved):

rest/learn_client.go
liujed: Oh, I entirely missed the intended behaviour of this function. Yeah, it's no longer possible to get an unaggregated timeline. Changed aggregation to 1-minute and fixed the function name in 671427a.

liujed:
Re-requesting review for changes in 671427a.

PR #133, opened 2022-06-28
Print packet capture stats after N seconds

thatplguy:
Adds a new flag "--stats-log-delay N" to apidump and learn.  After N seconds
(default 60), prints packet capture statistics + warnings/errors based on
packet capture behavior so far.

This ensures that users deploying the Akita CLI as a long-running process can
see capture statistics close to when the command was started, which is useful
for debugging.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -18,7 +18,6 @@ import (
 	"github.com/akitasoftware/akita-cli/ci"
 	"github.com/akitasoftware/akita-cli/deployment"
 	"github.com/akitasoftware/akita-cli/location"
-	"github.com/akitasoftware/akita-cli/pcap"
 	"github.com/akitasoftware/akita-cli/plugin"
 	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/rest"
@@ -96,6 +95,9 @@ type Args struct {
 
 	// Deployment tag value; may come from an environment variable
 	Deployment string
+
+	// Print packet capture statistics after N seconds.
+	StatsLogDelay int
 }
 
 func (args *Args) lint() {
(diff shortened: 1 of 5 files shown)


apidump/apidump.go
```diff
@@ -399,6 +338,30 @@ func Run(args Args) error {
...
 	var toRotate []trace.LearnSessionCollector
 
+	dumpSummary := NewSummary(
+		capturingNegation,
+		interfaces,
+		negationFilters,
+		numUserFilters,
+		filterSummary,
+		prefilterSummary,
+		negationSummary,
```
liujed, line 348: These summaries are read at the same time as they are being updated by the live packet capture. Without synchronization between the two threads, there is no guarantee that the reader will see any writes.

apidump/summary.go
```diff
@@ -0,0 +1,177 @@
...
+				printer.Stderr.Infof("%s\n", printer.Color.Yellow("This may mean your filter is incorrect, such as the wrong TCP port."))
+			}
+		} else if totalCount.Unparsed > 0 {
+			printer.Stderr.Infof("Captured %d TCP packets total; %d unparsed TCP segments.\n",
+				totalCount.TCPPackets, totalCount.Unparsed)
+			printer.Stderr.Infof("%s\n", printer.Color.Yellow("This may mean you are trying to capture HTTPS traffic."))
+			printer.Stderr.Infof("See https://docs.akita.software/docs/proxy-for-encrypted-traffic\n")
+			printer.Stderr.Infof("for instructions on using a proxy, or generate a HAR file with\n")
+			printer.Stderr.Infof("your browser as described in\n")
+			printer.Stderr.Infof("https://docs.akita.software/docs/collect-client-side-traffic-2\n")
```
liujed, line 93: Is this outdated?

liujed (commented):
Some concerns about the lack of concurrency control on the capture summaries.

apidump/apidump.go
thatplguy: Under the hood, these summaries use a `sync.RWMutex`.  The reader can't acquire a new read lock if there is a pending write lock, which should ensure that any pending writes preceding the stats dump are resolved before the stats dump can proceed.  Is there something there I'm missing?

apidump/summary.go
thatplguy: This is still valid, although the docs could use some updated screenshots.

apidump/apidump.go
liujed: Ah, you're right. Sorry I missed this!

liujed (approved):

apidump/summary.go
mgritter, line 93: This is OK as console output, but it's less good as logging output where it's split across multiple lines.  If output format is JSON, then at least some of these lines should be combined.

mgritter (approved):

apidump/summary.go
thatplguy: Hm, good point.  I'll consolidate lines.

PR #134, opened 2022-06-30
Send debugging telemetry to the backend

thatplguy:
Depends on https://github.com/akitasoftware/akita-libs/pull/142.

### DIFF
--- a/rest/learn_client.go
+++ b/rest/learn_client.go
@@ -190,6 +190,17 @@ func (c *learnClientImpl) GetSpecDiffTrie(ctx context.Context, baseID, newID aki
 	return &resp, err
 }
 
+func (c *learnClientImpl) PostClientPacketCaptureStats(ctx context.Context, serviceID akid.ServiceID, deployment string, req kgxapi.PostClientPacketCaptureStatsRequest) error {
+	// TODO(cns): Where is this normally set?
+	if deployment == "" {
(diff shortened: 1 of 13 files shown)


mgritter (approved):
I sort of expected that telemetry stats would be updated on a regular basis, rather than just once after 30 seconds?

liujed (approved):

rest/learn_client.go
```diff
@@ -190,6 +190,17 @@ func (c *learnClientImpl) GetSpecDiffTrie(ctx context.Context, baseID, newID aki
 	return &resp, err
 }
 
+func (c *learnClientImpl) PostClientPacketCaptureStats(ctx context.Context, serviceID akid.ServiceID, deployment string, req kgxapi.PostClientPacketCaptureStatsRequest) error {
+	// TODO(cns): Where is this normally set?
+	if deployment == "" {
+		deployment = "default"
+	}
+
+	path := fmt.Sprintf("/v1/services/%s/telemetry/%s/client", akid.String(serviceID), deployment)
```
liujed, line 199: AKIDs have `String` methods now, so the `akid.String` should no longer be necessary.

thatplguy:
This PR just focuses on whether the client was healthy on startup.  I'll look at continuous telemetry in a future PR.

PR #135, opened 2022-07-02
Bugfix: Telemetry summaries had too many entries

thatplguy:
Fixes a bug where too many entries were added to telemetry summaries if the
entries had the same value for TCPPackets.

### DIFF
--- a/trace/stats.go
+++ b/trace/stats.go
@@ -147,6 +147,9 @@ type pair[T constraints.Ordered] struct {
 	v *PacketCounts
 }
 
+// Return a new map with the N entries in counts with the highest TCP packet
+// counts.  In the case of a tie for the Nth position, the entry with the
+// smallest key is selected.
 func topNByTcpPacketCount[T constraints.Ordered](counts map[T]*PacketCounts, n int) map[T]*PacketCounts {
 	rv := make(map[T]*PacketCounts, math.Min(len(counts), n))
 
--- a/trace/stats.go
+++ b/trace/stats.go
@@ -8,6 +8,7 @@ import (
 	"github.com/akitasoftware/go-utils/math"
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
+	"golang.org/x/exp/constraints"
 	"golang.org/x/exp/slices"
 )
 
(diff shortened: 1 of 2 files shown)


trace/stats.go
```diff
@@ -141,37 +142,30 @@ func (s *PacketCounter) Summary(n int) *PacketCountSummary {
...
-		tcpPacketCounts = append(tcpPacketCounts, c.TCPPackets)
-	}
-	slices.SortFunc(tcpPacketCounts, func(a, b int) bool {
-		return b < a
+	// Sort descending by TCPPackets.
+	slices.SortFunc(pairs, func(a, b pair[T]) bool {
+		if a.v.TCPPackets != b.v.TCPPackets {
+			return b.v.TCPPackets < a.v.TCPPackets
+		}
+		return a.k < b.k
```
mgritter, line 163: I was a bit confused due to the tiebreaker being in the opposite order as the normal check.

trace/stats.go
```diff
@@ -141,37 +142,30 @@ func (s *PacketCounter) Summary(n int) *PacketCountSummary {
...
 }
 
-func topNByTcpPacketCount[T comparable](counts map[T]*PacketCounts, n int) map[T]*PacketCounts {
-	if n <= 0 {
-		return map[T]*PacketCounts{}
-	}
+type pair[T constraints.Ordered] struct {
+	k T
+	v *PacketCounts
+}
```
mgritter, line 148: I wondered if this could be put inside the function, but it cannot: "type declarations inside generic functions are not currently supported"?!

mgritter (approved):

trace/stats.go
```diff
@@ -141,37 +142,30 @@ func (s *PacketCounter) Summary(n int) *PacketCountSummary {
...
 
+func topNByTcpPacketCount[T constraints.Ordered](counts map[T]*PacketCounts, n int) map[T]*PacketCounts {
```
liujed, line 150: Please add a comment to specify expected behaviour. (The case of having ties did occur to me, but I figured you really did want to return them all.)

liujed (approved):

trace/stats.go
thatplguy: It's an arbitrary stable tiebreaker.  Choosing the larger value would have been fine as well.

trace/stats.go
thatplguy: I know!  I tried that first, but no luck 😞

PR #137, opened 2022-07-09
Send telemetry for the entire duration of the trace.

mgritter:
Upload telemetry every 5 minutes while apidump is running.

I did not reset the counters; this, along with https://github.com/akitasoftware/superstar/pull/903, means that the packet count for the entire lifetime of the client will be shown.  (If we want to show a timeline or other derived data on the UI, we can compute the deltas between samples there.)

I did not enable this for `learn`, but it would not be hard to add the (hidden) flag there too.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -43,6 +43,12 @@ const (
 	// processing.
 	// We budget for 5x to be safe.
 	pcapStopWaitTime = 5 * time.Second
+
+	// Number of top ports to show in telemetry
+	topNForSummary = 10
+
+	// Context timeout for telemetry upload
+	telemetryTimeout = 30 * time.Second
 )
 
 const (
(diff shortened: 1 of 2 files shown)


thatplguy (approved):

apidump/apidump.go
```diff
@@ -100,6 +100,9 @@ type Args struct {
 
 	// Print packet capture statistics after N seconds.
 	StatsLogDelay int
+
+	// Periodically report telemetry every N seconds thereafter
+	TelemetryInterval int
```
liujed, line 105: Why not use a `time.Duration`? (Seems like `StatsLogDelay` should also have been a `time.Duration`.)

apidump/apidump.go
```diff
@@ -226,6 +229,67 @@ func RotateLearnSession(args *Args, done <-chan struct{}, collectors []trace.Lea
...
+	req := kgxapi.PostClientPacketCaptureStatsRequest{
+		ClientID:                  args.ClientID,
+		ObservedStartingAt:        time.Now().UTC(),
+		ObservedDurationInSeconds: args.StatsLogDelay,
+	}
+
+	// Upload "req" to the server.
+	send := func() {
+		req.PacketCountSummary = dumpSummary.FilterSummary.Summary(10)
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
liujed, line 254: Suggest using symbolic constants

liujed (approved):

apidump/apidump.go
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/137/commits/0e73da19024c7bcd9cc2c274e302e4f46b91fd00

PR #139, opened 2022-07-11
Refactor pcap out of trace

thatplguy:
Factor out methods in `trace` that depend on the pcap library.  This makes it possible to
compile other CLI functionality (e.g. processing HAR files) without linking
against libpcap.

As part of this, I'm removing `learn/learn_runner.go`, which appears to be unused 
but was causing a circular dependency.


liujed (approved):
Thanks!

mgritter (approved):

PR #140, opened 2022-07-14
Continue packet collection after collector errors

thatplguy:
Unless all collectors have failed, continue collecting traffic on healthy collectors/interfaces.

As an example, ephemeral interfaces are sometimes removed while the Akita client is collecting traffic, e.g. if a Docker container is stopped.  Rather than kill the CLI, we now continue collecting and summarize errors when packet collection is stopped.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -664,10 +664,12 @@ func Run(args Args) error {
 					printer.Stderr.Infof("Received %v, stopping trace collection...\n", received.String())
 					break DoneWaitingForSignal
 				case interfaceErr := <-errChan:
-					printer.Stderr.Errorf("Encountered an error on interface %s, continuing with remaining interfaces.  Error: %s\n", interfaceErr.interfaceName, interfaceErr.err.Error())
 					errorsByInterface[interfaceErr.interfaceName] = interfaceErr.err
 
-					if len(errorsByInterface) == numCollectors {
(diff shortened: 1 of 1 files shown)


apidump/apidump.go
```diff
@@ -631,12 +655,22 @@ func Run(args Args) error {
...
+
+			// Continue until an interrupt or all collectors have stopped with errors.
+		DoneWaitingForSignal:
+			for {
+				select {
+				case received := <-sig:
+					printer.Stderr.Infof("Received %v, stopping trace collection...\n", received.String())
+					break DoneWaitingForSignal
+				case interfaceErr := <-errChan:
+					printer.Stderr.Errorf("Encountered an error on interface %s, continuing with remaining interfaces.  Error: %s\n", interfaceErr.interfaceName, interfaceErr.err.Error())
```
liujed, line 667: This will print even when there are no remaining interfaces, which might be confusing. I don't think this is a blocker for the PR, though.

liujed (approved):

apidump/apidump.go
thatplguy: Good idea, fixed in [7fd0f98](https://github.com/akitasoftware/akita-cli/pull/140/commits/7fd0f988c09858ee5c7225564ca7153d9e154cfe).

PR #147, opened 2022-08-20
Memory-usage improvements

liujed:
Two improvements to memory usage:
* Don't buffer HTTP requests and responses during parsing. Previously, we were buffering the entire request/response so that it could be emitted as raw data when parsing failed. Instead we now just emit a count of how much data was consumed.
* Read HTTP payloads from a fixed-size buffer pool. Currently, this is a 20 MiB pool that gets allocated in 4-KiB chunks. These numbers were chosen arbitrarily.

I am concerned that this second change will cause the agent to stop reading full HTTP payloads when there are a lot of concurrent connections.

To evaluate the effects of these changes, 10 concurrent GET requests were continually made to a test service, once every <20 seconds for at least 6 hours. The response payload consisted of about 27 MiB of random JSON data. As a result of these changes, the agent's maximum RSS as reported by Linux (VmHWM) was reduced by 76%, from 1,307 MiB to 310 MiB. The effect of the first change was also measured independently: VmHWM was reduced by 41%, from 1,307 MiB to 771 MiB.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -370,6 +370,11 @@ func Run(args Args) error {
 		}
 	}
 
+	// If --dogfood is specified, enable assertions in the buffer-pool code.
+	if viper.GetBool("dogfood") {
+		buffer_pool.CheckInvariants = true
+	}
+
 	// Create a buffer pool for storing HTTP payloads.
 	pool, err := buffer_pool.MakeBufferPool(20*1024*1024, 4*1024)
 	if err != nil {
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	kgxapi "github.com/akitasoftware/akita-libs/api_schema"
+	"github.com/akitasoftware/akita-libs/buffer_pool"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
 
@@ -369,6 +370,12 @@ func Run(args Args) error {
 		}
 	}
 
+	// Create a buffer pool for storing HTTP payloads.
+	pool, err := buffer_pool.MakeBufferPool(20*1024*1024, 4*1024)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to create buffer pool")
+	}
+
 	// If the output is targeted at the backend, create a shared backend
 	// learn session.
 	var backendSvc akid.ServiceID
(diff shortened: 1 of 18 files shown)


liujed:
Depends on akitasoftware/akita-libs#150.

apidump/apidump.go
```diff
@@ -369,6 +370,12 @@ func Run(args Args) error {
 		}
 	}
 
+	// Create a buffer pool for storing HTTP payloads.
+	pool, err := buffer_pool.MakeBufferPool(20*1024*1024, 4*1024)
```
mgritter, line 374: I would like to see some back-of-the-envelope estimate on how much this limits our collection rate.

Flickr is 380,000 events per day = 260 per second.  Little's law says that if they spend 40ms resident then that is 10.4 requests resident a time, call it 12.  That is a generous 1.8MB per request.

But, we actually throw a lot of Flickr's requests away after parsing them!  Their true data rate may be much higher; we could look at the telemetry to figure out how much... if they were successfully sending telemetry, dangit.

If we're throwing away 90% of their requests, that changes the allocation to 200KB per request.  Still seems OK.

mgritter: The other thing to consider is how much this increases our fixed overhead, and whether we should reduce the GC parameter to compensate.

20MB of live data = 40MB of RSS footprint

If our footprint used to be 400MB, it's now 440MB.  But we could get that back by setting GOGCLIMIT=81 instead.  (180MB + 220MB live heap = 400MB, 220 * 0.81 =~ 180

pcap/run.go
```diff
@@ -35,7 +36,9 @@ func Collect(stop <-chan struct{}, intf, bpfFilter string, bufferShare float32,
 
 	for t := range parsedChan {
 		t.Interface = intf
-		if err := proc.Process(t); err != nil {
+		err := proc.Process(t)
+		t.Content.ReleaseBuffers()
```
mgritter, line 40: Hm, I was misunderstanding this in my akita-libs review, there's no need to worry about different readers of the buffer having different lifetimes.

But, that does suggest that we maybe need not actually allocate the buffers from a fixed pool; we only need synchronize on how much memory is used.  I don't think this is a blocker; I don't want you to rewrite it.  But I do want to put these on the record in case we need to pursue further performance tuning.

In a multi-CPU system the allocator can be thread- or CPU-local which avoids thrashing cache lines.  With the shared pool we're giving up any potential optimization here.  (I know tcmalloc does exactly this sort of thing.  What the Go runtime does, I have no idea.)

Keeping the buffer live has a cost in footprint and in future profiling efforts.  It also turns an accounting issue (early release of the buffer before reading is done) into a correctness issue (re-use of old data.)

That said, if this is causing spikes then maybe re-use is a win because the GC isn't having to wake up.  So let's definitely keep it as-is, but maybe we want to consider making our buffers "virtual" rather than "real".

mgritter (approved):
Looks good.  Thanks for including your testing results in the PR!

pcap/run.go
liujed: Yeah, my primary goal was to have a memory pool to avoid GC altogether. Caching issues did occur to me, but I figured we could start with something simple first and dig deeper in the future, if desired.

PR #151, opened 2022-08-31
Preserve JSON-based custom content types

thatplguy:
Previously, the CLI handled JSON-encoded custom content types by replacing the
custom type with 'application/json'.  This PR changes that behavior to preserve
the name of the custom content type while still treating it as JSON.

### DIFF
--- a/learn/parse_http.go
+++ b/learn/parse_http.go
@@ -289,13 +289,15 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
 		return nil, errors.Wrapf(err, "failed to parse MIME from Content-Type %q", contentType)
 	}
 
-	// Rewrite media type to JSON for types encoded as JSON
+	// Handle media types encoded as JSON by parsing as JSON and recording the
+	// media type as "other".
 	// TODO: XML parsing
 	// TODO: application/json-seq (RFC 7466)?
 	// TODO: more text/* types
+	isMediaJson := mediaType == "application/json"
 	switch {
 	case strings.HasSuffix(mediaType, "+json"):
-		mediaType = "application/json"
+		isMediaJson = true
 	}
 
 	var bodyData *pb.Data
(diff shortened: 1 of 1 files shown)


learn/parse_http.go
```diff
@@ -326,66 +328,74 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
 		bodyData = parseElem(string(body), interpret)
 	}
 
-	switch mediaType {
-	case "application/json":
+	// Handle JSON media types.
+	if isMediaJson {
 		bodyData, err = parseHTTPBodyJSON(bodyStream)
```
mgritter, line 333: It's too bad we can't mix conditions and value-matching in a single switch statement.

One way to do this would be instead of the boolean isMediaJson, have a string parseAsMediaType which starts as mediaType and is changed in the JSON case.  This would handle cases like text/csv too, since they could be parsed as text/plain instead of adding another boolean to handle that.

learn/parse_http.go
```diff
@@ -326,66 +328,74 @@ func parseBody(contentType string, bodyStream io.Reader, statusCode int) (*pb.Da
...
+			return parseMultipartBody("form-data", mediaParams["boundary"], bodyStream, statusCode)
+		case "multipart/mixed":
+			return parseMultipartBody("mixed", mediaParams["boundary"], bodyStream, statusCode)
+		case "application/octet-stream":
+			handleAsBlob()
+			pbContentType = pb.HTTPBody_OCTET_STREAM
+		case "text/plain", "text/csv":
+			// If the text is just a number, report its type
+			handleAsString(spec_util.INTERPRET_STRINGS)
+			pbContentType = pb.HTTPBody_TEXT_PLAIN
```
mgritter, line 391: We're doing the same thing for non-JSON types here, do we want to start recording the original content type in "OtherType" all the time?

I think that is probably the right way to go, even though it makes the field name incorrect.

mgritter (approved):
This is OK to go, but while we're here we should think about preserving the original media type in all cases (text/csv, text/x-yaml, etc.)

learn/parse_http.go
thatplguy: Good idea.  Added in 278330d4e0d2a9c6ea226e9b1704c9e4aa144d4e.

mgritter (approved):

learn/parse_http.go
thatplguy: I changed this method to use `pb.HTTPBody_ContentType` to indicate how to parse the body.  Now the flow is (a) interpret `mediaType` to set `parseBodyDataAs`, and (b) switch on `parseBodyDataAs` to do the parsing.

PR #155, opened 2022-09-20
Add Segment telemetry to CLI.

mgritter:
Moved GetClientID() to telemetry package, and Domain to the rest package, to avoid accessing the "internal" package, which is forbidden by the compiler.

Attempts to look up /v1/user for user identity, with several fallback options.

Has an opt-out mechanism, and rate limit for frequent errors.

Depends on https://github.com/akitasoftware/akita-libs/pull/154

The Segment key must be compiled in to official builds, or provided as an environment variable.


mgritter:
Verified that there are events showing up in Segment in the 'akita-cli' source, but they have apk_XXX id's because /v1/user doesn't return anything if you call it with basic auth.

rest/models.go
```diff
@@ -2,9 +2,13 @@ package rest
 
 import (
 	"github.com/akitasoftware/akita-libs/akid"
+	"github.com/akitasoftware/akita-libs/api_schema"
 )
 
+// TODO: shouldn't this be in akita-cli/api_schema?
```
liujed, line 8: Did you intend to address this?

telemetry/telemetry.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		printer.Stderr.Errorf("Telemetry unavailable; error setting up Segment client: %v\n", err)
+		analyticsClient = nullClient{}
+	}
+}
+
+func getDistinctID() string {
+	// If we have a user email, use that!
+	// Otherwise use the configured API Key.
+	// Failing that, try to the user name and host name?
```
liujed, line 94:
```suggestion
	// Failing that, try to use the user name and host name?
```

telemetry/telemetry.go
```diff
@@ -0,0 +1,259 @@
...
+	id := os.Getenv("AKITA_SEGMENT_DISTINCT_ID")
+	if id != "" {
+		return id
+	}
+
+	// TODO: every command creates its client ID separately. That makes it
+	// impossible to use the same one here without major refactoring.
+	// So we use the all-zeros ID so it's obviously suspect.
+	frontClient := rest.NewFrontClient(rest.Domain, GetClientID())
+	userResponse, err := frontClient.GetUser(context.Background())
```
liujed, line 105: Should we use a timeout here?

pcap/stream.go
```diff
@@ -152,6 +153,9 @@ func (f *tcpFlow) reassembledWithIgnore(ignoreCount int, sg reassembly.ScatterGa
 
 		f.currentParser = nil
 		f.currentParserCtx = nil
+
+		// FIXME; is this going to be too high data-rate?
+		telemetry.Error("parser", err)
```
liujed, line 158: Did you intend to address this? Looks like `telemetry.RateLimitError` might be an alternative.

liujed (approved):
Looks good. Might also be worth reporting select environment variables, so we can get visibility into OS, architecture, Docker, Kubernetes, etc.

rest/models.go
mgritter: No, not really, I didn't take the time to investigate why this was the only model here.

telemetry/telemetry.go
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/155/commits/15edda28ca047f11bdd13abe447c1b651785fb49

pcap/stream.go
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/155/commits/bafd054e083b8a473730d71ca4b58ff43abd12af

mgritter:
I think reporting on platform information has to be its own task.  I looked somewhat into grabbing environment variable names (we want to avoid sending values, I think) but there's no standard ones that would tell us that we're in Docker, Kubernetes, GCE, etc.  There are things we could do but they would be more involved.

PR #156, opened 2022-09-21
Bugfix: Duplicate data items

thatplguy:
HTTP requests/responses may contain elements that result in duplicate Data items in the IR.  For example, requests may contain cookies with the same name and value (but different domains or paths, which is not currently stored in the IR).

Previously, duplicate items were interpreted as hash collisions, causing parsing to fail.  This PR amends that behavior to check for equality and fail only if items are different.

### DIFF
--- a/learn/parse_http.go
+++ b/learn/parse_http.go
@@ -163,10 +163,15 @@ func ParseHTTP(elem akinet.ParsedNetworkContent) (*PartialWitness, error) {
 	for _, d := range datas {
 		// Use the hash of the data proto as the key so we can deterministically
 		// compare witnesses.
+		//
+		// Some witnesses contain duplicate data elements, such as cookies with
+		// the same name and value (but different domains or paths, which we
+		// don't currently include in the IR).
+		//
+		// If there is a potential hash collision, fall back to an expensive
+		// equality check.  If the witnesses are different, return an error.
 		k := ir_hash.HashDataToString(d)
 		if existing, collision := dataMap[k]; collision && !proto.Equal(d, existing) {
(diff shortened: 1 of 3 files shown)


learn/parse_http.go
```diff
@@ -162,8 +163,15 @@ func ParseHTTP(elem akinet.ParsedNetworkContent) (*PartialWitness, error) {
...
+		// Some witnesses contain duplicate data elements, such as cookies with
+		// the same name and value (but different domains or paths, which we
+		// don't currently include in the IR).
+		//
+		// If there is a potential hash collision, fall back to an expensive
+		// equality check.  If the witnesses are different, return an error.
 		k := ir_hash.HashDataToString(d)
-		if _, collision := dataMap[k]; collision {
+		if existing, collision := dataMap[k]; collision && !proto.Equal(d, existing) {
 			return nil, errors.Errorf("detected collision in data map key")
```
mgritter, line 175: I'm not sure failing to record the witness is correct anyway (but at least with the new telemetry we'll be able to tell.)  The alternative is to include just an arbitrary one of the hashed data objects, or do a comparison that makes it deterministic even if unequal.

It would probably be better if we had a "evil bit" in the IR to record that there were some errors in parsing, and so we maybe shouldn't use it for the model, but still include in the timeline.

mgritter (approved):

learn/parse_http.go
thatplguy: I thought about that.  Making an arbitrary choice would probably work in aggregate, but in small numbers there's a risk of producing models that flip flop over time.

The "evil bit" idea sounds good to me.  Another option would be to add an extra field to Data objects that the CLI can assign random values to in order to break hash collisions, although this would increase the size of deserialized IR objects.

PR #162, opened 2022-09-23
Send telemetry about startup errors to the backend.

mgritter:
Made a start at refactoring apidump.Run() into smaller methods.
Update akita-libs.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -2,7 +2,6 @@ package apidump
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


apidump/apidump.go
```diff
@@ -121,6 +122,106 @@ type Args struct {
...
+// Send the initial mesage to the backend indicating successful start
+func (a *apidump) SendInitialTelemetry() {
+	// The observed duration serves as a key for upsert, so
+	// it should be the same on the initial empty report indicating
+	// succesful startup, and the one sixty seconds later.
+	req := &kgxapi.PostClientPacketCaptureStatsRequest{
+		ObservedDurationInSeconds: a.StatsLogDelay,
+	}
+	a.SendTelemetry(req)
+}
```
liujed, line 175: The old version of this had `req.PacketCountSummary = dumpSummary.FilterSummary.Summary(topNForSummary)`. Is this no longer needed/desired?

liujed (approved):

apidump/apidump.go
mgritter: The initial summary should be empty.  I puzzled about that for a while; the API is documented such that PacketCountSummary should be absent, so I was surprised to see the call there. 

> 	// If PacketCountSummary is absent, then this observation period has
>     // started but not yet concluded.
 
 This seems to work OK on the server side, but it is a behavior change.

PR #165, opened 2022-10-20
Discard control characters when parsing JSON

thatplguy:
The Go JSON parser rejects JSON strings that contain control characters (U+0000
- U+001F).  We'd like to be more permissive.  Rather than implementing our own parser, this PR adds a preprocessing step that removes unescaped control characters before invoking Go's JSON parser.

### DIFF
--- a/learn/json_preprocessor.go
+++ b/learn/json_preprocessor.go
@@ -1,0 +1,54 @@
+package learn
+
+import (
+	"io"
+	"unicode"
+)
+
+type stripControlCharactersReader struct {
+	wrapped io.Reader
+}
+
+func newStripControlCharactersReader(wrapped io.Reader) stripControlCharactersReader {
+	return stripControlCharactersReader{wrapped: wrapped}
+}
+
+// Read up to len(p) bytes, removing any unescaped control characters found.
+// Removed characters do not count toward the total bytes read.  Returns the
+// number of bytes read.
+func (r stripControlCharactersReader) Read(p []byte) (n int, err error) {
+	pIdx := 0
+
+	// Read up to len(p) bytes, then discard any control characters.  Continue
+	// reading (and discarding control characters) until p is full or there are
+	// no more bytes to read.
+	for pIdx < len(p) {
+		remaining := len(p) - pIdx
+		buf := make([]byte, remaining)
+
+		var bufN int
+		bufN, err = r.wrapped.Read(buf)
+
+		// Copy from buf to p, skipping unescaped control characters.
+		hasPrecedingSlash := false
+		for _, r := range string(buf) {
+			if unicode.IsControl(r) && !hasPrecedingSlash {
+				continue
+			}
+			if r == '\\' {
+				hasPrecedingSlash = !hasPrecedingSlash
+			}
+			pIdx += copy(p[pIdx:], string([]rune{r}))
+		}
+
+		// If we hit an error or read fewer bytes than the size of the buffer,
+		// don't bother trying to read more from the underlying reader.
+		if err != nil || bufN < remaining {
+			break
+		}
+
+		// Otherwise, we loop to replace CRs we dropped.
+	}
+
+	return pIdx, err
+}
--- a/learn/json_preprocessor.go
+++ b/learn/json_preprocessor.go
@@ -2,7 +2,6 @@ package learn
 
 import (
 	"io"
-	"unicode"
 )
 
 type stripControlCharactersReader struct {
@@ -13,7 +12,7 @@ func newStripControlCharactersReader(wrapped io.Reader) stripControlCharactersRe
 	return stripControlCharactersReader{wrapped: wrapped}
 }
 
-// Read up to len(p) bytes, removing any unescaped control characters found.
+// Read up to len(p) bytes, removing any control characters found.
 // Removed characters do not count toward the total bytes read.  Returns the
 // number of bytes read.
 func (r stripControlCharactersReader) Read(p []byte) (n int, err error) {
@@ -30,12 +29,13 @@ func (r stripControlCharactersReader) Read(p []byte) (n int, err error) {
 		var bufN int
 		bufN, err = r.wrapped.Read(bufSlice)
 
-		// Copy from buf to p, skipping unescaped control characters.
-		for _, r := range string(bufSlice) {
-			if unicode.IsControl(r) {
+		// Copy from buf to p, skipping control characters.
+		for _, c := range bufSlice {
+			if c < 0x1f {
 				continue
 			}
-			pIdx += copy(p[pIdx:], string([]rune{r}))
+			p[pIdx] = c
+			pIdx += 1
 		}
 
 		// If we hit an error or read fewer bytes than the size of the buffer,
--- a/learn/json_preprocessor.go
+++ b/learn/json_preprocessor.go
@@ -18,19 +18,20 @@ func newStripControlCharactersReader(wrapped io.Reader) stripControlCharactersRe
 // number of bytes read.
 func (r stripControlCharactersReader) Read(p []byte) (n int, err error) {
 	pIdx := 0
+	buf := make([]byte, len(p))
 
 	// Read up to len(p) bytes, then discard any control characters.  Continue
 	// reading (and discarding control characters) until p is full or there are
 	// no more bytes to read.
 	for pIdx < len(p) {
 		remaining := len(p) - pIdx
-		buf := make([]byte, remaining)
+		bufSlice := buf[:remaining]
 
 		var bufN int
-		bufN, err = r.wrapped.Read(buf)
+		bufN, err = r.wrapped.Read(bufSlice)
 
 		// Copy from buf to p, skipping unescaped control characters.
-		for _, r := range string(buf) {
+		for _, r := range string(bufSlice) {
 			if unicode.IsControl(r) {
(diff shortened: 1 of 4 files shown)


learn/json_preprocessor.go
```diff
@@ -0,0 +1,54 @@
...
+
+		var bufN int
+		bufN, err = r.wrapped.Read(buf)
+
+		// Copy from buf to p, skipping unescaped control characters.
+		hasPrecedingSlash := false
+		for _, r := range string(buf) {
+			if unicode.IsControl(r) && !hasPrecedingSlash {
+				continue
+			}
```
mgritter, line 37: I'm a bit puzzled by this, the sequence '\' , '\n' is not valid either.  I don't think we need to pay attention to \ to make this work.

If there's a case I'm not getting, can you put a test case in for it?

learn/json_preprocessor.go
```diff
@@ -0,0 +1,54 @@
...
+	// no more bytes to read.
+	for pIdx < len(p) {
+		remaining := len(p) - pIdx
+		buf := make([]byte, remaining)
+
+		var bufN int
+		bufN, err = r.wrapped.Read(buf)
+
+		// Copy from buf to p, skipping unescaped control characters.
+		hasPrecedingSlash := false
```
liujed, line 33: The value of this should be persisted in `stripControlCharactersReader`. Not only can the last character in the previous loop be a backslash, so can be the last character that was read from the stream during the last call to `Read`.

learn/json_preprocessor.go
```diff
@@ -0,0 +1,54 @@
...
+				hasPrecedingSlash = !hasPrecedingSlash
+			}
+			pIdx += copy(p[pIdx:], string([]rune{r}))
+		}
+
+		// If we hit an error or read fewer bytes than the size of the buffer,
+		// don't bother trying to read more from the underlying reader.
+		if err != nil || bufN < remaining {
+			break
+		}
```
liujed, line 48: Should this error also invalidate future calls to `Read`?

learn/json_preprocessor.go
liujed: I think the idea is to avoid turning the closed string `"\<CR>"` into the open string `"\"`.

liujed: Come to think of it, instead of preserving escaped control characters, it may be better to discard both the control character and its escaping backslash, so that `"\<CR>"` turns into the empty string instead of being kept.

thatplguy: Hm, good point.  We want to avoid removing escaped control characters, e.g. `"foo\\r"` should be left alone, but I suppose that's an issue when encoding rather than decoding -- by the time we parse it, the rune is either a control character or not.

I [added a field](https://github.com/akitasoftware/akita-cli/blob/0ed1dd0a60f7c5b66fce570a5f0475ee721b1e24/learn/parse_http_test.go#L68) with an escaped CRLF to the JSON object we use in the unit tests, but the test passes after removing the `hasPrecedingSlash` logic above.

Fixed in fea612b431db1793c20b496fc9c34c66556efafc.

liujed: I think you wanted `"foo\\\r"` instead of `"foo\\r"`.

learn/json_preprocessor.go
```diff
@@ -0,0 +1,54 @@
...
+// number of bytes read.
+func (r stripControlCharactersReader) Read(p []byte) (n int, err error) {
+	pIdx := 0
+
+	// Read up to len(p) bytes, then discard any control characters.  Continue
+	// reading (and discarding control characters) until p is full or there are
+	// no more bytes to read.
+	for pIdx < len(p) {
+		remaining := len(p) - pIdx
+		buf := make([]byte, remaining)
```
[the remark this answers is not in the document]
thatplguy: The `Read` interface uses the size of the buffer as an indicator of how much to read.  But you're right, rather than allocating a new buffer, I can use a slice over an existing buffer to manage the size.  Changed in b3361deaed335f184d152808a5dbb039472298bf.

learn/json_preprocessor.go
thatplguy: Removed this in fea612b431db1793c20b496fc9c34c66556efafc.

liujed: Can you try with the following JSON?
```
{
  "greeting": "hello \
",
  "subject": "world"
}
```

learn/json_preprocessor.go
```diff
@@ -0,0 +1,54 @@
...
+		// Copy from buf to p, skipping unescaped control characters.
+		hasPrecedingSlash := false
+		for _, r := range string(buf) {
+			if unicode.IsControl(r) && !hasPrecedingSlash {
+				continue
+			}
+			if r == '\\' {
+				hasPrecedingSlash = !hasPrecedingSlash
+			}
+			pIdx += copy(p[pIdx:], string([]rune{r}))
```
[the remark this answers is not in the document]
thatplguy: I don't think that's the case.  We only read up to the remaining free bytes in `p`.  If anything, that might cause us to read the first byte of a multi-byte character and then decode it improperly, as well as any future bytes read.  

If we get very unlucky, the first byte could decode to a control character, which we would then remove, invalidating the rest of the string.  I think that would cause the parser to fail though, which isn't any worse than what we're currently doing.

learn/json_preprocessor.go
thatplguy: This passes through errors from the wrapped reader.  I don't think we need to do anything special to invalidate future calls; the wrapped reader will presumably be invalidated and return an error in the future.

learn/json_preprocessor.go
liujed: Right, which is why I suggested that we also remove the backslash: https://github.com/akitasoftware/akita-cli/pull/165#discussion_r1000090469.

learn/json_preprocessor.go
thatplguy: Sorry @liujed, didn't see your comment until I refreshed the page.

I was mistaken -- the JSON parser doesn't support escaping control characters in strings.  E.g. `"foo\\\r"` fails to parse.  You can write `"foo\\r"`, which escapes the slash and avoids the control character entirely.

learn/json_preprocessor.go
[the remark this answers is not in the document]
liujed: As discussed offline, bytes in multi-byte UTF-8 code points all have their high bits set, so this isn't a concern.

learn/json_preprocessor.go
```diff
@@ -0,0 +1,51 @@
...
+	// reading (and discarding control characters) until p is full or there are
+	// no more bytes to read.
+	for pIdx < len(p) {
+		remaining := len(p) - pIdx
+		bufSlice := buf[:remaining]
+
+		var bufN int
+		bufN, err = r.wrapped.Read(bufSlice)
+
+		// Copy from buf to p, skipping unescaped control characters.
```
liujed, line 33:
```suggestion
		// Copy from buf to p, skipping control characters.
```

learn/json_preprocessor.go
```diff
@@ -0,0 +1,51 @@
...
+		for _, r := range string(bufSlice) {
+			if unicode.IsControl(r) {
+				continue
+			}
```
liujed, line 37: Seems like we could do without the `[]byte`-to-`string` conversion (which is a copy!). Instead, we could iterate over the `[]byte` and check whether `r <= 0x1f`.

liujed (approved):

learn/json_preprocessor.go
thatplguy: Good call, fixed in 451cae0a07f2548db5b0267ff872ef900ef33c88.

learn/json_preprocessor.go
thatplguy: Fixed in 451cae0a07f2548db5b0267ff872ef900ef33c88.

learn/json_preprocessor.go
```diff
@@ -0,0 +1,51 @@
...
+
+type stripControlCharactersReader struct {
+	wrapped io.Reader
+}
+
+func newStripControlCharactersReader(wrapped io.Reader) stripControlCharactersReader {
+	return stripControlCharactersReader{wrapped: wrapped}
+}
+
+// Read up to len(p) bytes, removing any unescaped control characters found.
```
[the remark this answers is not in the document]
thatplguy: Fixed in 451cae0a07f2548db5b0267ff872ef900ef33c88.

learn/json_preprocessor.go
thatplguy: As discussed offline, we're going to leave this as is for now.  It's unlikely to result in successful but mangled parsing except in pathological cases, and it addresses a situation that isn't valid JSON in the first place.

learn/json_preprocessor.go
```diff
@@ -0,0 +1,51 @@
...
+	// no more bytes to read.
+	for pIdx < len(p) {
+		remaining := len(p) - pIdx
+		bufSlice := buf[:remaining]
+
+		var bufN int
+		bufN, err = r.wrapped.Read(bufSlice)
+
+		// Copy from buf to p, skipping control characters.
+		for _, c := range bufSlice {
```
liujed, line 33:
```suggestion
		for _, c := range bufSlice[:bufN] {
```

PR #170, opened 2022-11-03
Explain permission errors in a context-sensitve way.

mgritter:
Only suggest "use sudo" if not running as root.
There's also some customized messages if the interface list is empty, which... I don't really expect is possible?

### DIFF
--- a/apidump/net.go
+++ b/apidump/net.go
@@ -42,7 +42,7 @@ func showPermissionErrors(sampleError error) error {
 		if os.Geteuid() == 0 {
 			if env.InDocker() {
 				printer.Warningf("Although you are running as root, this container lacks the CAP_NET_RAW capability.\n")
-				printer.Warningf("It might be that you are in a PaaS which disallows packet capture, or the local configuration has disabled that privilege by default.\n")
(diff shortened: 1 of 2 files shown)


apidump/net.go
```diff
@@ -29,6 +31,40 @@ func (w interfaceWrapper) Addrs() ([]net.Addr, error) {
...
+func showPermissionErrors(sampleError error) error {
+	// TODO: pcapPermError needs an Unwrap method to use error.Is().  But
+	// I don't know what error this is anyway! So we're back to substring search.
+	if strings.Contains(sampleError.Error(), "Operation not permitted") {
+		// Permission denied == not enough capabilities
+		// Are we running as root?
+		if os.Geteuid() == 0 {
+			if env.InDocker() {
+				printer.Warningf("Although you are running as root, this container lacks the CAP_NET_RAW capability.\n")
+				printer.Warningf("It might be that you are in a PaaS which disallows packet capture, or the local configuration has disabled that privilege by default.\n")
```
liujed, line 45: Not all PaaS's disallow packet capture.
```suggestion
				printer.Warningf("It might be that you are in a PaaS that disallows packet capture, or the local configuration has disabled that privilege by default.\n")
```

liujed (approved):

PR #173, opened 2022-11-05
Refactor batching in backend collector

liujed:
Processing of witnesses now occurs when they are enqueued for upload. Additionally, the size of the buffer is now based on memory usage, rather than number of items.


liujed:
Depends on akitasoftware/akita-libs#170.

PR #174, opened 2022-11-05
Don't send witnesses that are too large

liujed:
Depends on #173.


apidump/apidump.go
```diff
@@ -604,13 +608,13 @@ func (a *apidump) Run() error {
...
-					backendCollector = trace.NewBackendCollector(a.backendSvc, backendLrn, a.learnClient, args.Plugins)
+					backendCollector = trace.NewBackendCollector(a.backendSvc, backendLrn, a.learnClient, optionals.Some(a.MaxWitnessSize_bytes), args.Plugins)
 					collector = trace.TeeCollector{
 						Dst1: backendCollector,
 						Dst2: localCollector,
 					}
 				} else if args.Out.AkitaURI != nil {
-					backendCollector = trace.NewBackendCollector(a.backendSvc, backendLrn, a.learnClient, args.Plugins)
+					backendCollector = trace.NewBackendCollector(a.backendSvc, backendLrn, a.learnClient, optionals.Some(a.MaxWitnessSize_bytes), args.Plugins)
 					collector = backendCollector
```
mgritter, line 618: I don't recall why we have a BackendCollector per interface, it seems like we are buffering in more than once place.

If we are serious about reducing the memory used to a max of 30MB total, then maybe we should revisit this.

trace/report_buffer.go
```diff
@@ -39,6 +42,8 @@ func (buf *reportBuffer) Add(raw rawReport) (bool, error) {
 		witnessReport, err := raw.Witness.toReport()
 		if err != nil {
 			printer.Warningf("Failed to convert witness to report: %v\n", err)
+		} else if maxSize, exists := buf.maxWitnessSize_bytes.Get(); exists && len(witnessReport.WitnessProto) > maxSize {
+			// Drop the witness; it's too large.
```
mgritter, line 46: Should at least have a debug statement here; you've already discussed adding a count to packet telemetry.
(Which might be _more_ difficult if the BackendCollector and its reportBuffer is not associated with a particular interface any more.)

mgritter (approved):

trace/report_buffer.go
liujed: Added debug message in 65fa486.

PR #175, opened 2022-11-08
add-to-ECS: select profile, region, and cluster

mgritter:
This covers:
  * asking for a profile name
  * selecting a region from a list
  * listing the clusters in a region


mgritter:
Example interaction:
```
mark@ubuntu:~/akita-cli$ bin/akita ecs  
? Which of your AWS profiles should Akita use to configure ECS? mgritter
[INFO] Successfully loaded AWS credentials.
? In which AWS region is your ECS cluster? us-west-2
[ERROR] Could not find any ECS clusters in this region. Please select a different one or hit Ctrl+C to exit.
? In which AWS region is your ECS cluster?  [Use arrows to move, type to filter, ? for more help]
> default
  find all clusters
  af-south-1
  ap-east-1
  ap-northeast-1
  ap-northeast-2
  ap-northeast-3
[INFO] Interrupted!
```

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,330 @@
...
+	"us-west-2",
+}
+
+// Ask the user to select a region.
+func getRegionState(wf *AddWorkflow) (keepGoing bool, err error) {
+	reportStep("Get AWS Region")
+
+	if awsRegionFlag != "" {
+		wf.awsRegion = awsRegionFlag
+		err = wf.createSession()
```
mgritter, line 175: I have set this up so that if command-line arguments are provided, we do not prompt for that step at all.

The intent is that specifying all the command-line flags should be the same whether or not the program is running interactively. However, an alternative is that in interactive mode we could use the flags as the default and ask for confirmation.  I'm still undecided, so feedback about when we ask for confirmation would be welcome: only at the end?  At each step?  Not at all?

go.mod
```diff
@@ -3,14 +3,15 @@ module github.com/akitasoftware/akita-cli
...
-	github.com/AlecAivazis/survey/v2 v2.3.2
+	github.com/AlecAivazis/survey/v2 v2.3.6
 	github.com/OneOfOne/xxhash v1.2.8
 	github.com/Pallinder/go-randomdata v1.2.0
 	github.com/akitasoftware/akita-ir v0.0.0-20220630210013-8926783978fe
 	github.com/akitasoftware/akita-libs v0.0.0-20221101223658-d0dd96f228c7
 	github.com/akitasoftware/go-utils v0.0.0-20220606224752-aad0f81bb9e7
 	github.com/akitasoftware/plugin-flickr v0.2.0
 	github.com/andybalholm/brotli v1.0.1
+	github.com/aws/aws-sdk-go v1.44.132
```
liujed, line 14: Why not use aws-sdk-go-v2?

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,330 @@
...
+	"eu-west-2",
+	"eu-west-3",
+	"me-central-1",
+	"me-south-1",
+	"sa-east-1",
+	"us-east-1",
+	"us-east-2",
+	"us-west-1",
+	"us-west-2",
+}
```
liujed, line 167: Surely, we can get this list programmatically, so that we don't have to maintain it.

liujed (commented):
I'll look at this in more detail later, but here are a couple of quick comments.

cmd/internal/ecs/add.go
mgritter: That's what I said!  But I haven't found it yet. Maybe it's in one of the services.

I'll take another look or do some google searches.

liujed: This looks promising: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.DescribeRegions

cmd/internal/ecs/add.go
liujed: How about if we waited until the end to confirm, but provided a command-line flag to skip this final confirmation?

mgritter:
Converting to draft until sdk v2 conversion is complete.

liujed (commented):
A few more comments. I'll do the rest of the review when the change to SDK v2 lands.

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,330 @@
...
+	wf.awsCred = credentials.NewSharedCredentials(awsCredentialsFlag, wf.awsProfile)
+	if err = wf.checkCredentials(); err != nil {
+		printer.Errorf("Error from AWS library: %v\n", err)
+		printer.Errorf("Could not find AWS credentials for profile %q, please try again or hit Ctrl+C to exit.\n", wf.awsProfile)
+		wf.awsProfile = ""
+		return true, nil
+	}
+
+	// TODO: I don't think there's even a way to show the profile name here?
+	printer.Infof("Successfully loaded AWS credentials.\n")
```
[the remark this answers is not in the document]
mgritter: I was confused because I thought the default profile could be something other than "default", and so there was no way to load the actual value if we used "".

But, I was wrong on the first point, and the new version uses "default" as the default profile name instead of "", as the latter doesn't work with the new config anyway.

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,330 @@
...
+
+// Helper function for reporting telemetry
+func reportStep(stepName string) {
+	telemetry.WorkflowStep("Add to ECS", stepName)
+}
+
+// A function which executes the next part of the workflow,
+// and picks a next state or exits. Return "true" to continue
+// so that the zero value exits.
+type AddWorkflowState func(*AddWorkflow) (cont bool, err error)
```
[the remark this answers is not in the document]
mgritter: This sounded totally reasonable, but:

```
cmd/internal/ecs/add.go:93:9: cannot use optionals.Some(getProfileState) (value of type optionals.Optional[func(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowState], err error)]) as type optionals.Optional[AddWorkflowState] in return statement
```

This only works if either the type is an alias, or every function literal is wrapped in a AddWorkflowState constructor.  :(

It might be better to forgo the Optional here and use nil as the end value, which I don't love.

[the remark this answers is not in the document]
mgritter: Added some helpers to make this work less painfully.

go.mod
mgritter: Done!

cmd/internal/ecs/add.go
mgritter: I left the list as a fallback, because it is entirely possible not to have DescribeRegions permissions.

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+	}
+
+	choices := make([]string, 0, len(arnToName))
+	for c, _ := range arnToName {
+		choices = append(choices, string(c))
+	}
+	sort.Slice(choices, func(i, j int) bool {
+		return choices[i] < choices[j]
+	})
```
liujed, line 267:
```suggestion
	sort.Strings(choices)
```

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,390 @@
...
+	err = survey.AskOne(
+		&survey.Select{
+			Message: "In which cluster does your application run?",
+			Help:    "Select ECS cluster with the task you want to modify.",
+			Options: choices,
+			Description: func(value string, _ int) string {
+				name := arnToName[arn(value)]
+				if name == "" {
+					return ""
+				}
```
liujed, line 279: When does this happen?

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,390 @@
...
+				return name + " in " + arnToRegion[arn(value)]
+			},
+		},
+		&clusterAnswer,
+	)
+
+	wf.ecsClusterARN = arn(clusterAnswer)
+	wf.createClient(arnToRegion[wf.ecsClusterARN])
+	wf.awsRegion = arnToRegion[wf.ecsClusterARN]
+	wf.ecsCluster = arnToName[wf.ecsClusterARN]
```
liujed, line 289:
```suggestion
	wf.ecsCluster = arnToName[wf.ecsClusterARN]
	wf.awsRegion = arnToRegion[wf.ecsClusterARN]
	wf.createClient(wf.awsRegion)
```

cmd/internal/ecs/add.go
```diff
@@ -0,0 +1,390 @@
...
+		return awf_next(getRegionState)
+	}
+
+	printer.Infof("Found %d clusters in region %q.\n", len(clusters), wf.awsRegion)
+
+	choices := make([]string, 0, len(clusters))
+	for c, _ := range clusters {
+		choices = append(choices, string(c))
+	}
+	choices = append(choices, goBackOption)
```
liujed, line 328:
```suggestion
	choices := make([]string, 0, len(clusters))
	for c, _ := range clusters {
		choices = append(choices, string(c))
	}
	choices = append(choices, goBackOption)
	sort.Strings(choices)
```

cmd/internal/ecs/aws_api.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+// Create an ECS client with the specified region
+func (wf *AddWorkflow) createClient(region string) {
+	wf.awsConfig.Region = region
+	wf.ecsClient = ecs.NewFromConfig(wf.awsConfig)
+}
+
+// Create an ECS client asuming that the config has a default region.
```
liujed, line 105:
```suggestion
// Create an ECS client assuming that the config has a default region.
```

cmd/internal/ecs/aws_api.go
```diff
@@ -0,0 +1,221 @@
...
+	"me-south-1",
+	"sa-east-1",
+	"us-east-1",
+	"us-east-2",
+	"us-west-1",
+	"us-west-2",
+}
+
+// List all regions. On error fall back to the precanned list.
+func (wf *AddWorkflow) listAWSRegions() []string {
```
liujed, line 138:
```suggestion
// List all regions in alphabetical order. On error fall back to the precanned list.
func (wf *AddWorkflow) listAWSRegions() (result []string) {
	defer func() { sort.Strings(result) }()
```

liujed (approved):

liujed:
GitHub appears confused.
image

liujed (approved):

cmd/internal/ecs/add.go
mgritter: The cluster may not have a name.

cmd/internal/ecs/add.go
mgritter: I'll make this choice but the goBackOption should not be sorted.

cmd/internal/ecs/ecs.go
```diff
@@ -0,0 +1,78 @@
...
+	Short:        Cmd.Short,
+	Long:         Cmd.Long,
+	SilenceUsage: true,
+	RunE:         addAgentToECS,
+}
+
+var RemoveFromECSCmd = &cobra.Command{
+	Use:          "remove",
+	Short:        "Remove the Akita agent from AWS ECS.",
+	Long:         "Remove a previously-installed Akita container from an ECS Task.",
```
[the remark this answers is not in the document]
mgritter: I disagree on this one but convention seems to be on your side.  :(

cmd/internal/ecs/add.go
mgritter: Fixed in https://github.com/akitasoftware/akita-cli/pull/175/commits/15c83ddcd47ed5309971ccd116768b1b1c2776bb

PR #176, opened 2022-11-08
Move default definitions

liujed:
Moved the definitions of default command-line parameter values out of `apidump` and into `apispec`, to avoid having `apispec` depend on `apidump`. This allows client code to import `apispec` without having to also pull in pcap.

### DIFF
--- a/apispec/defaults.go
+++ b/apispec/defaults.go
@@ -1,4 +1,4 @@
-package apidump
+package apispec
 
 import "time"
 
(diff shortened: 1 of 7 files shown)


mgritter (approved):
This is fine, but I wish you had put them in some third location because apispec may go away in the future.

liujed:
I had the same concern, but they can be moved easily enough if/when that time comes. I originally tried putting them in the `cmd` package, but that introduced a dependency cycle.

PR #177, opened 2022-11-09
Detect when packet capture fails from architecture mismatch

thatplguy:
When running the Akita agent in a Docker container built for a different architecture than the host (e.g. built for amd64 but running on arm64), the agent fails to read any network interfaces with error messages of the form `SIOCETHTOOL(ETHTOOL_GET_TS_INFO) ioctl failed: Function not implemented`.

This PR extends the agent to detect these errors and print a nicer error message for the user that indicates the underlying problem.  It also adds the target architecture to the `akita --version` output, e.g. `akita version 0.21.18 (09c7847085262f10ea9e2cd4bae280383970e5c1-dirty, arm64)`.

When running in Docker, the CLI output will look like this:
```
> docker run -e AKITA_API_KEY_ID -e AKITA_API_KEY_SECRET akitasoftware/cli:latest apidump --project akibox

WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
[INFO] Akita Agent 0.21.18-rc4
[WARNING] Skipping interface eth0 for collecting packets because of error: failed to read packets from interface eth0: eth0: SIOCETHTOOL(ETHTOOL_GET_TS_INFO) ioctl failed: Function not implemented
[WARNING] Skipping interface lo for collecting packets because of error: failed to read packets from interface lo: lo: SIOCETHTOOL(ETHTOOL_GET_TS_INFO) ioctl failed: Function not implemented
[WARNING] The agent received "Function not implemented" when trying to read from your network interfaces. This often indicates that the Akita agent was built for a different architecture than your host architecture. This Akita agent binary was built for arm64.
[ERROR] No network interfaces could be used: Unable to read network interfaces. If your host architecture is not arm64, try using `docker pull --platform $YOUR_ARCHITECTURE akitasoftware/cli:latest` to pull an Akita agent built for your architecture.
```

When not in Docker, the last error message will instead read:
```
[ERROR] No network interfaces could be used: Unable to read network interfaces.  If your host architecture is not arm64, try downloading an Akita agent built for your architecture from https://github.com/akitasoftware/akita-cli/releases.
```

### Testing
I tested this by building a CLI docker image on arm64 and running it on an amd64 macOS laptop.

### DIFF
--- a/apidump/net.go
+++ b/apidump/net.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -13,6 +12,7 @@ import (
 	"github.com/google/gopacket/pcap"
 	"github.com/pkg/errors"
 
+	"github.com/akitasoftware/akita-cli/architecture"
 	"github.com/akitasoftware/akita-cli/env"
 	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/telemetry"
@@ -58,17 +58,7 @@ func showPermissionErrors(sampleError error) error {
 	} else if strings.Contains(sampleError.Error(), "SIOCETHTOOL(ETHTOOL_GET_TS_INFO) ioctl failed: Function not implemented") {
 		// This happens when the binary was built for a different architecture, e.g.
 		// if the user pulled the amd64 Docker image on arm64.
-
-		// Use architecture names consistent with our CLI releases.
-		var arch string
-		switch runtime.GOARCH {
-		case "x86_64", "amd64":
-			arch = "amd64"
-		case "aarch64", "arm64":
-			arch = "arm64"
-		default:
-			arch = runtime.GOARCH
-		}
+		arch := architecture.GetCanonicalArch()
 
 		printer.Warningf(
 			"The agent received \"Function not implemented\" when trying to read from your network interfaces. "+
--- a/apidump/net.go
+++ b/apidump/net.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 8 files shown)


apidump/net.go
```diff
@@ -54,12 +55,36 @@ func showPermissionErrors(sampleError error) error {
...
+		arch := architecture.GetCanonicalArch()
+
+		printer.Warningf(
+			"The agent received \"Function not implemented\" when trying to read from your network interfaces. "+
+				"This often indicates that the Akita agent was built for a different architecture than your host architecture. "+
+				"This Akita agent binary was built for %s.\n",
+			arch,
+		)
+
+		if os.Getenv("__X_AKITA_CLI_DOCKER") == "true" {
```
liujed, line 70:
```suggestion
		if env.InDocker() {
```

liujed:
I started reviewing this, but GitHub is returning 500s when I try to add comments to my review. I'll come back to this later.

apidump/net.go
```diff
@@ -54,12 +55,36 @@ func showPermissionErrors(sampleError error) error {
...
+			return errors.Errorf(
```
mgritter, line 71: Do we want to distinguish this in the client telemetry?

https://github.com/akitasoftware/akita-cli/blob/d94be84ef8c80a2c1de3b73b9e704bf2af755471/apidump/apidump.go#L431-L436

Right now it is listed as a permission error but we could add a new cause to our enum (and then handle it on the backend...)  Our Segment events will have the raw errors, so we'll see it either way, but the user can't if looking at the UI.  (I don't recall exactly what they see, do we show the original error message or just one based on type?)

apidump/net.go
```diff
@@ -54,12 +55,36 @@ func showPermissionErrors(sampleError error) error {
...
+				"Unable to read network interfaces. If your host architecture is not %s, try using "+
+					"`docker pull --platform $YOUR_ARCHITECTURE akitasoftware/cli:latest` to pull an Akita agent "+""+
```
mgritter, line 73: Why the spare `+""+`?

apidump/net.go
```diff
@@ -54,12 +55,36 @@ func showPermissionErrors(sampleError error) error {
...
+					"built for your architecture.",
+				arch,
+			)
+		} else {
+			return errors.Errorf(
+				"Unable to read network interfaces. If your host architecture is not %s, try downloading an Akita agent built for your architecture from https://github.com/akitasoftware/akita-cli/releases.",
```
liujed, line 79: GitHub only has releases for Linux. They should be using the install script if they're not running one of our container images.

liujed (approved):

mgritter (approved):
Looks good.

apidump/net.go
mgritter: This is what we show for that error type, it would be good to define a new type or pick OTHER, which does display the original.
```
        return (
          <Warning
            title="Client Lacks Permissions"
            message={`The Akita client could not capture traffic because it lacks permissions to do so. In a container environment, ensure that the container which has the client can run in privileged mode and has the CAP_NET_RAW capability; this might be disabled by default, or not permitted by your service provider. If running directly on a host or VM, start the client as the root user, or set the CAP_NET_RAW capability on the binary. Please contact support@akitasoftware.com with a description of your runtime environment if you need further assistance.`}
          />
        );
```

apidump/net.go
thatplguy: Fixed in 0b00a81e70aa176948cf99f1dbef4fba1b00a01e.

apidump/net.go
thatplguy: Good point, fixed in 0b00a81e70aa176948cf99f1dbef4fba1b00a01e.

apidump/net.go
thatplguy: Hm, maybe IntelliJ autocorrected a typo.  Fixed in 0b00a81e70aa176948cf99f1dbef4fba1b00a01e.

apidump/net.go
thatplguy: Good idea.  I'm going to leave this PR open while I add support for a new telemetry error to the backend, and then add that here.

thatplguy:
Depends on https://github.com/akitasoftware/akita-libs/pull/173.

thatplguy:
Requesting rereview for b5c9f3f2db465cc382b869857459f1d5d03db398.

apidump/net.go
```diff
@@ -43,26 +45,51 @@ func showPermissionErrors(sampleError error) error {
...
+
+		printer.Warningf(
+			"The agent received \"Function not implemented\" when trying to read from your network interfaces. "+
+				"This often indicates that the Akita agent was built for a different architecture than your host architecture. "+
+				"This Akita agent binary was built for %s.\n",
+			arch,
+		)
+
+		if env.InDocker() {
+			return errors.Errorf(
```
mgritter, line 72: Missing NewApidumpError here?  Will take a longer look later.

thatplguy: Whoops, thanks.  Fixed in a4cf4764fb2a61494fbbb60af14620c089ede27e.

mgritter (approved):
Looks good.

PR #178, opened 2022-11-09
Count oversized witnesses and report to back end

liujed:
Add and maintain a count of oversized witnesses in `client_telemetry.PacketCounts`. This is not reported to the user, but is reported to the back end as part of packet telemetry.


liujed:
Depends on akitasoftware/akita-libs#172.

thatplguy (approved):

PR #182, opened 2022-11-15
Add health check utility for Docker Extension

versilis:
This adds a server to the `apidump` command which responds to health check requests sent from the Docker extension.

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -125,6 +125,11 @@ type Args struct {
 
 	// The maximum witness size to upload. Anything larger is dropped.
 	MaxWitnessSize_bytes int
+
+	// Whether to run the command with additional functionality to support the Docker Extension
+	DockerExtensionMode bool
+	// The port to be used by the Docker Extension for health checks
+	HealthCheckPort int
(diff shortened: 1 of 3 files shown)


apidump/apidump.go
```diff
@@ -404,12 +409,36 @@ type interfaceError struct {
...
+
+	go func() {
+		errChan <- startHealthCheckServer(args.HealthCheckPort)
+	}()
+
+	go func() {
+		errChan <- run(args)
+	}()
+
+	return <-errChan
```
liujed, line 439: Let's unify these.
```suggestion
	errChan := make(chan error)

	// The Docker extension expects a health-check server to be running. Only
	// start this server if it's needed.
	if args.DockerExtensionMode {
		go func() {
			errChan <- startHealthCheckServer(args.HealthCheckPort)
		}()
	}

	// Run the main packet-capture loop.
	go func() {
		args.lint()

		a := newSession(&args)
		errChan <- a.Run()
	}()

	return <-errChan
```

liujed (approved):

PR #184, opened 2022-11-17
Create AWS secrets to hold Akita credentials

mgritter:
...if they do not already exist.

Check that credentials are present and project exists.

### DIFF
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -89,8 +89,6 @@ const (
 // from the other command conventions, but there are relatively few
 // usage errors here.)
 func RunAddWorkflow() error {
-	// Verify that we have credentials to the Akita CLI
-
 	wf := &AddWorkflow{
 		currentState: initState,
 		ctx:          context.Background(),
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -135,9 +135,9 @@ func RunAddWorkflow() error {
 //    |    ^  |
 //    |    |  V
 //    |-getService
-//           |
-//           V
-//     checkForSecret
+//    |      |
+//    |      V
+//    |- getSecret
 //           |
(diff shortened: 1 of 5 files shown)


cmd/internal/ecs/add.go
```diff
@@ -571,10 +589,38 @@ func getServiceState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
 	return awf_next(confirmState)
```
liujed, line 589:
```suggestion
	return awf_next(getSecretState)
```

cmd/internal/ecs/add.go
```diff
@@ -122,6 +137,9 @@ func RunAddWorkflow() error {
 //    |-getService
 //           |
 //           V
+//     checkForSecret
```
liujed, line 140:
```suggestion
//    |-getService
//    |      |
//    |      V
//    |- getSecret
```

cmd/internal/ecs/add.go
```diff
@@ -664,5 +715,42 @@ func fillFromFlags(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowSta
 }
 
 func addSecretState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowState], err error) {
-	return awf_error(errors.New("Unimplemented!"))
+	reportStep("Add AWS Secret")
+
+	key, secret := cfg.GetAPIKeyAndSecret()
```
liujed, line 720: Would be nice if we requested a new key and secret from the back end for use in ECS. Not a blocker, though.

liujed (approved):

cmd/internal/ecs/add.go
mgritter: I am very confused as to how this worked when I tested it.

... oh, because they both default to false.

PR #185, opened 2022-11-18
Register a modified ECS task definition.

mgritter:
Check whether the task definition already contains an Akita container, or the tag we will add.
The code to create an AWS secret is disabled for now, because using the secret requires changes to the assigned role.
Recognize ECS-specific environment variables and tag the trace.

Depends on https://github.com/akitasoftware/akita-libs/pull/177 for the new tags.

Verified that the new task definition works by upgrading the service in the ECS console.

### DIFF
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -62,25 +62,36 @@ type AddWorkflow struct {
 	ecsClusterARN arn
 
 	ecsTaskDefinitionFamily string
+	ecsTaskDefinitionTags   []types.Tag
 	ecsTaskDefinitionARN    arn
 	ecsTaskDefinition       *types.TaskDefinition
 
 	ecsService    string
 	ecsServiceARN arn
 
-	akitaSecrets secretState
+	// TODO: provide a flag that re-enables use of secrets
+	// The problem is that the container's assumed role needs permission to
+	// read the configured secrets, which seems difficult to set up here.
+	secretsEnabled bool
+	akitaSecrets   secretState
 }
 
 const (
 	// Tag to use for objects created by the Akita CLI
-	akitaCreationTagKey   = "akita.software:created_by"
-	akitaCreationTagValue = "Akita Software ECS integration"
+	akitaCreationTagKey       = "akita.software:created_by"
+	akitaCreationTagValue     = "Akita Software ECS integration"
+	akitaModificationTagKey   = "akita.software:modified_by"
+	akitaModificationTagValue = "Akita Software ECS integration"
 
 	// Separate AWS secrets for the key ID and key secret
 	// TODO: make these configurable
 	akitaSecretPrefix    = "akita.software/"
 	defaultKeyIDName     = akitaSecretPrefix + "api_key_id"
 	defaultKeySecretName = akitaSecretPrefix + "api_key_secret"
+
+	// Akita agent image locations
+	akitaECRImage    = "public.ecr.aws/akitasoftware/akita-cli"
+	akitaDockerImage = "akitasoftware/cli"
 )
 
 // Run the "add to ECS" workflow until we complete or get an error.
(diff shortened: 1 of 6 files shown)


mgritter:
The trace created during testing does not have the new tags because it's running the existing container image, but it does have the AWS region tag:

```
chrome-ocelot-218b60a0         2022-11-18T06:00:03Z
                               x-akita-aws-region=us-east-2
                               x-akita-cli-arch=amd64
                               x-akita-cli-env=docker
                               x-akita-cli-os=linux
                               x-akita-cli-version=0.21.20
                               x-akita-deployment=default
                               x-akita-source=deployment
```

thatplguy (approved):

cmd/internal/ecs/add.go
```diff
@@ -752,5 +797,79 @@ func addSecretState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowSt
...
+	}
+
+	input.Tags = append(input.Tags, types.Tag{
+		Key:   aws.String(akitaCreationTagKey),
+		Value: aws.String(akitaCreationTagValue),
+	})
+
+	apiKey, apiSecret := cfg.GetAPIKeyAndSecret()
+	input.ContainerDefinitions = append(input.ContainerDefinitions, types.ContainerDefinition{
+		Name: aws.String("akita"),
```
liujed, line 835:
```suggestion
		Name: aws.String("akita-agent"),
```

cmd/internal/ecs/add.go
```diff
@@ -752,5 +797,79 @@ func addSecretState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowSt
...
+		// TODO: Cpu and Memory should be omitted for Fargate; they take their default values for EC2 if omitted.
+		// For now we can leave the defaults in place, but they might be a bit large for EC2.
+		EntryPoint: []string{"/akita", "apidump", "--project", projectFlag},
+		Environment: []types.KeyValuePair{
+			// Setting these environment variables will cause the traces to be tagged.
+			{Name: aws.String("AKITA_AWS_REGION"), Value: &wf.awsRegion},
+			{Name: aws.String("AKITA_ECS_SERVICE"), Value: &wf.ecsService},
+			{Name: aws.String("AKITA_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
+			{Name: aws.String("AKITA_API_KEY_ID"), Value: &apiKey},
+			{Name: aws.String("AKITA_API_KEY_SECRET"), Value: &apiSecret},
```
liujed, line 845:
```suggestion
		Environment: []types.KeyValuePair{
			{Name: aws.String("AKITA_API_KEY_ID"), Value: &apiKey},
			{Name: aws.String("AKITA_API_KEY_SECRET"), Value: &apiSecret},
			
			// Setting these environment variables will cause the traces to be tagged.
			{Name: aws.String("AKITA_AWS_REGION"), Value: &wf.awsRegion},
			{Name: aws.String("AKITA_ECS_SERVICE"), Value: &wf.ecsService},
			{Name: aws.String("AKITA_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
```

cmd/internal/ecs/add.go
```diff
@@ -752,5 +797,79 @@ func addSecretState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowSt
...
+		},
+		Essential: aws.Bool(false),
+		Image:     aws.String(akitaECRImage),
+		Secrets:   []types.Secret{},
```
liujed, line 849: Can this be omitted? Or is the intention for there to be an explicit placeholder in case we end up using Secrets Manager? If the latter, let's say so in a comment.

liujed (approved):

cmd/internal/ecs/add.go
mgritter: Yeah, I'll remove it.

PR #186, opened 2022-11-18
Update ECS service and wait for deployment to complete.

mgritter:
Handle config files in a docker volume named /aws.
Un-hide 'ecs' subcommand.
Added utility function for converting 'arn' type to *string for AWS SDK.

### DIFF
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -901,6 +901,7 @@ func updateServiceState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkfl
 	}
 	_, err = wf.ecsClient.UpdateService(wf.ctx, input)
 	if err != nil {
+		telemetry.Error("AWS ECS UpdateService", err)
 		if uoe, unauth := isUnauthorized(err); unauth {
 			printer.Errorf("The provided credentials do not have permission to update the ECS service %q (operation %s).\n",
 				wf.ecsServiceARN, uoe.OperationName)
@@ -922,6 +923,7 @@ func updateServiceState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkfl
 	if tagErr == nil {
 		printer.Infof("Tagged service %q with %q.\n", wf.ecsService, akitaModificationTagKey)
 	} else {
+		telemetry.Error("AWS ECS TagResource", err)
 		if uoe, unauth := isUnauthorized(tagErr); unauth {
 			printer.Warningf("The provided credentials do not have permission to tag the ECS service %q (operation %s).\n",
 				wf.ecsServiceARN, uoe.OperationName)
@@ -965,6 +967,7 @@ func waitForRestartState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkf
 		select {
 		case t := <-ticker.C:
 			if t.After(endTime) {
+				reportStep("EC2 Service Deployment Timeout")
 				printer.Warningf("Giving up after five minutes.\n")
 				if deployment.RunningCount > 0 {
 					printer.Infof("Some tasks did start, indicating that the new task definition is OK.\n")
@@ -984,6 +987,7 @@ func waitForRestartState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkf
 		}
 	}
 	if deployment.RolloutState == types.DeploymentRolloutStateFailed {
+		telemetry.Error("EC2 Deployment", errors.New(aws.ToString(deployment.RolloutStateReason)))
 		printer.Errorf("Deployment of the new task definition failed. The reason given by AWS is: %q",
 			aws.ToString(deployment.RolloutStateReason))
 		// TODO: guide the user through this?
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/internal/ecs/add.go
```diff
@@ -870,6 +879,121 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
...
+				// TODO: guide the user through this?
+				printer.Infof("You can use the AWS web console to delete the task definition \"%s:%s\". The previous task definition should still be in use.\n",
+					wf.ecsTaskDefinitionFamily, wf.ecsTaskDefinition.Revision)
+				return awf_error(errors.New("The modified task failed to deploy. Please contact support@akitasoftware.com for assistance."))
+			}
+		}
+	}
+
+	reportStep("ECS Service Updated")
+	printer.Infof("Deployment successful! Please return to the Akita web console.\n")
```
liujed, line 997: Provide a link?

cmd/internal/ecs/add.go
```diff
@@ -870,6 +879,121 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
...
+		return awf_error(failure)
+	}
+	printer.Infof("Found new deployment with ID %q; its state is %s.\n", deploymentID, deployment.RolloutState)
+
+	if deployment.RolloutState == types.DeploymentRolloutStateInProgress {
+		printer.Infof("Waiting for the deployment to reach COMPLETED.\n")
+	}
+
+	ticker := time.NewTicker(15 * time.Second)
+	for deployment.RolloutState == types.DeploymentRolloutStateInProgress {
```
thatplguy, line 964: I think there's a race condition where `deployment.RolloutState == types.DeploymentRolloutStateFailed` before reaching this loop, in which case the CLI will report success despite the failure.

Maybe break on line 985 and check/report on failure after the loop?

cmd/internal/ecs/aws_api.go
```diff
@@ -87,8 +87,19 @@ func wrapUnauthorizedFor(err error, arn arn) error {
...
 func (wf *AddWorkflow) createConfig() error {
 	// Try loading credentials for the profile.
-	// TODO: specify a location inside the Docker container, if running on Docker
-	sharedConfig, err := config.LoadSharedConfigProfile(wf.ctx, wf.awsProfile)
+	// Use the default if running standalone.  Add /aws for use inside a container.
+	configFiles := []string{config.DefaultSharedConfigFilename(), "/aws/config"}
+	credentialsFiles := []string{config.DefaultSharedCredentialsFilename(), "/aws/credentials"}
+	if awsCredentialsFlag != "" {
+		credentialsFiles = append(credentialsFiles, awsCredentialsFlag)
+	}
```
liujed, line 95: Does the order here matter? We should make sure the user-specified files take precedence over the default files.

cmd/internal/ecs/aws_api.go
```diff
@@ -409,5 +429,38 @@ func (wf *AddWorkflow) getService(serviceARN arn) (*types.Service, error) {
 		return nil, fmt.Errorf("Mismatch between service and task definition; please choose a different task or service.")
 	}
 
-	return &svc, nil
+	return svc, nil
+}
+
+var noDeploymentFound = errors.New("No deployment found")
+
+func (wf *AddWorkflow) GetDeploymentMatchingTask(serviceARN arn) (string, types.Deployment, error) {
```
liujed, line 437:
```suggestion
// Returns the deployment of the given service that matches the ECS task
// definition configured in the workflow. For convenience, the deployment ID
// is also returned as a string.
func (wf *AddWorkflow) GetDeploymentMatchingTask(serviceARN arn) (string, types.Deployment, error) {
```

thatplguy (approved):
Looks good, aside from one corner case where we might miss reporting an error (see comment).

liujed (approved):

cmd/internal/ecs/add.go
mgritter: Good catch, thanks.  I don't think we need to break, the loop condition will terminate.

mgritter: in https://github.com/akitasoftware/akita-cli/pull/186/commits/e67a85920c734666661216bc194f17033a593ab5

cmd/internal/ecs/add.go
mgritter: I thought about it, but what we expect is that they have come from the onboarding workflow and just need to find the tab they had open.  If @paul-russo can suggest a route to use, though, I'd be happy to include one.

cmd/internal/ecs/aws_api.go
mgritter: I believe that's what it does, but the documentation complicates things:

> 	// If duplicate profiles are provided with a same, or across multiple
>	// shared credentials files, the next parsed profile will override only
>	// properties that conflict with the previously defined profile. Note that
>	// if duplicate profiles are provided within the SharedCredentialsFiles and
>	// SharedConfigFiles, the properties defined in shared credentials file
>	// take precedence.

I'm reading a lot into "next" but I believe that means it visits them in order and applies changes as it goes.

liujed: Gotcha, thanks. Your interpretation looks right.

cmd/internal/ecs/add.go
liujed: Makes sense. We can add a link later if it turns out that people are falling off here.

PR #191, opened 2022-12-31
Remove apispec and upload commands

liujed:
These are no longer used in any meaningful way. Code for the `upload` command remains, in case we wish to resurrect it later. The `apispec` command is replaced with a message directing the user to use `apidump` instead. But `upload` is completely removed, since there is no replacement.

```
$ bin/akita apispec
[INFO] Akita Agent 0.0.0
[ERROR] the apispec command has been removed. Please use apidump instead

$ bin/akita apispec --help
[INFO] Akita Agent 0.0.0
[ERROR] the apispec command has been removed. Please use apidump instead

$ bin/akita apispec foo bar baz
[INFO] Akita Agent 0.0.0
[ERROR] the apispec command has been removed. Please use apidump instead
```


mgritter (approved):

apispec/defaults.go
```diff
@@ -3,6 +3,10 @@ package apispec
 import "time"
 
 // Specifies default values for command-line parameters.
+//
+// These really ought to live in apidump, but that package depends on libpcap.
+// So we put these constants here to pulling in libpcap when importing these
+// constants.
```
liujed, line 9:
```suggestion
// So we put these constants here to avoid pulling in libpcap when importing
// these constants.
```

versilis (approved):

PR #195, opened 2023-01-30
Allow user to use a proxy, an alternate server name, or skip TLS

mgritter:
Would like to have a user who is blocked try this out.

Adds ability to specify an alternate server name, negative and positive testing:
```
mark@ubuntu:~/akita-cli$ sudo -E bin/akita apidump --server-tls-name api.akitasoftare.com --project mgg-test --debug
[INFO] Akita Agent 0.0.0
[ERROR]  request failed [error Get "https://api.akita.software/v1/services": x509: certificate is valid for api.akita.software, not api.akitasoftare.com method GET url https://api.akita.software/v1/services]

mark@ubuntu:~/akita-cli$ sudo -E bin/akita apidump --server-tls-name api.akita.software --project mgg-test --debug.
[INFO] Akita Agent 0.0.0
[DEBUG] Project name "mgg-test" is "svc_14kS9tQ5U2gS98De4TZIwc"
...
```

Adds configuration flag for proxy, accepts either URL or host:port.
```
mark@ubuntu:~/akita-cli$ sudo -E bin/akita apidump --proxy 173.243.116.9:55555 --project mgg-test --debug
[INFO] Telemetry unavailable; no Segment key configured.
[INFO] This is caused by building from source rather than using an official build.
[INFO] Akita Agent 0.0.0
[DEBUG] Using ID "apk_774YdN0HsvbPiwUtfpuPyR" for telemetry
[DEBUG]  CI not detected: `CI` is not set to `TRUE` in the environment (currently unset)
[DEBUG] Using proxy http://173.243.116.9:55555
[DEBUG] Project name "mgg-test" is "svc_14kS9tQ5U2gS98De4TZIwc"
...
[INFO] Created new trace on Akita Cloud: akita://mgg-test:trace:bold-lantern-3bd9ecf6
```

Also adds --skip-validate flag for bypassing TLS checks.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -178,10 +178,10 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&rest.Domain, "domain", defaultDomain, "Your assigned Akita domain (e.g. company.akita.software)")
 	rootCmd.PersistentFlags().MarkHidden("domain")
 
-	// Use a proxy or permit an unknown certificate.
-	rootCmd.PersistentFlags().StringVar(&rest.UseProxy, "proxy", "", "The domain name or IP address of an HTTP proxy server to use")
-	rootCmd.PersistentFlags().BoolVar(&rest.PermitInvalidCertificate, "skip-validate", false, "Skip TLS validation on the connection to api.akita.software")
-	rootCmd.PersistentFlags().MarkHidden("skip-validate")
+	// Use a proxy or permit a mismatched certificate.
+	rootCmd.PersistentFlags().StringVar(&rest.ProxyAddress, "proxy", "", "The domain name, IP address, or URL of an HTTP proxy server to use")
+	rootCmd.PersistentFlags().BoolVar(&rest.PermitInvalidCertificate, "skip-tls-validate", false, "Skip TLS validation on the connection to api.akita.software")
+	rootCmd.PersistentFlags().MarkHidden("skip-tls-validate")
 	rootCmd.PersistentFlags().StringVar(&rest.ExpectedServerName, "server-tls-name", "", "Provide an alternate TLS server name to accept instead of api.akita.software")
 	rootCmd.PersistentFlags().MarkHidden("server-tls-name")
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -178,6 +178,13 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&rest.Domain, "domain", defaultDomain, "Your assigned Akita domain (e.g. company.akita.software)")
 	rootCmd.PersistentFlags().MarkHidden("domain")
 
+	// Use a proxy or permit an unknown certificate.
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -178,6 +178,13 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&rest.Domain, "domain", defaultDomain, "Your assigned Akita domain (e.g. company.akita.software)")
 	rootCmd.PersistentFlags().MarkHidden("domain")
 
+	// Use a proxy or permit an unknown certificate.
+	rootCmd.PersistentFlags().StringVar(&rest.UseProxy, "proxy", "", "The domain name or IP address of an HTTP proxy server to use")
+	rootCmd.PersistentFlags().BoolVar(&rest.PermitInvalidCertificate, "skip-validate", false, "Skip TLS validation on the connection to api.akita.software")
+	rootCmd.PersistentFlags().MarkHidden("skip-validate")
```
liujed, line 184: Rename to `skip-tls-validate` or `skip-tls-validation`?

cmd/root.go
```diff
@@ -178,6 +178,13 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&rest.Domain, "domain", defaultDomain, "Your assigned Akita domain (e.g. company.akita.software)")
 	rootCmd.PersistentFlags().MarkHidden("domain")
 
+	// Use a proxy or permit an unknown certificate.
+	rootCmd.PersistentFlags().StringVar(&rest.UseProxy, "proxy", "", "The domain name or IP address of an HTTP proxy server to use")
```
liujed, line 182: It appears that a user can specify `https://proxy.example.net` or `socks5://proxy.example.net`, and the implementation will connect through the HTTPS or SOCKS5 proxy. Shall we document and support these too?

rest/base_client.go
```diff
@@ -22,6 +22,15 @@ import (
 // was buried in the "internal" package where we couldn't use it.
 var Domain string
 
+// Use a proxy, "" is none. (This is because the flags package doesn't support Optional)
+var UseProxy string
```
liujed, line 26: How about calling this `ProxyName` or `ProxyAddress` instead? If we are only supporting HTTP proxies, then this should also be indicated.

liujed (approved):

cmd/root.go
mgritter: Yes, my intention was to support any URI (though we don't validate that it is only one of HTTP, HTTPS, or SOCKS.)

I will change the comment to include the URI as an option, but I think of the service being named as an "HTTP proxy" independent of the protocol.

rest/base_client.go
mgritter: Changed in https://github.com/akitasoftware/akita-cli/pull/195/commits/3a73809e34f771b0dc7149c89e0b705710b66d1c

cmd/root.go
mgritter: Changed in https://github.com/akitasoftware/akita-cli/pull/195/commits/3a73809e34f771b0dc7149c89e0b705710b66d1c

versilis (approved):

PR #202, opened 2023-03-15
Add command to generate Kubernetes Secrets

versilis:
This adds a new command `akita kube secret` which generates a Kubernetes secret configuration file that stores a user's base-64 encoded Akita API credentials.

To simplify file generation, I've used go's built-in templating utilities; `akita-secret.tmpl` is used as the template for creating the secret.

Example usage:
```
% ./bin/akita kube secret -n test -o ./tmp/deployments/configs/akita-secrets.yml
[INFO] Akita Agent 0.0.0
[INFO] Generated Kubernetes secret config to ./tmp/deployments/configs/akita-secrets.yml                                                                                                               

% less ./tmp/deployments/configs/akita-secrets.yml 
apiVersion: v1
kind: Secret
metadata:
  name: akita-secrets
  namespace: test
type: Opaque
data:
  akita-api-key: YXBrXzN1Y2h6WDyiOdyMzg2NldiM3Y0Q2Y=
  akita-api-secret: YmUzMzY5OTllNzNjc2DY3MmI5OWQzMTVmAnGaKmYzN2NlNjc2NWRiZDY4MzNjMWRkMzA4YjFjZDFlNWZkZg==
./tmp/deployments/configs/akita-secrets.yml lines 1-9/9 (END)

### DIFF
--- a/cmd/internal/ecs/ecs.go
+++ b/cmd/internal/ecs/ecs.go
@@ -5,8 +5,6 @@ import (
 	"strings"
 
 	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
-	"github.com/akitasoftware/akita-cli/env"
-	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/rest"
 	"github.com/akitasoftware/akita-cli/telemetry"
 	"github.com/akitasoftware/akita-cli/util"
@@ -70,8 +68,18 @@ func init() {
 	Cmd.PersistentFlags().StringVar(&awsRegionFlag, "region", "", "The AWS region in which your ECS cluster resides.")
 	Cmd.PersistentFlags().StringVar(&ecsClusterFlag, "cluster", "", "The name or ARN of your ECS cluster.")
 	Cmd.PersistentFlags().StringVar(&ecsServiceFlag, "service", "", "The name or ARN of your ECS service.")
-	Cmd.PersistentFlags().StringVar(&ecsTaskDefinitionFlag, "task", "", "The name of your ECS task definition to modify.")
-	Cmd.PersistentFlags().BoolVar(&dryRunFlag, "dry-run", false, "Perform a dry run: show what will be done, but do not modify ECS.")
+	Cmd.PersistentFlags().StringVar(
+		&ecsTaskDefinitionFlag,
+		"task",
+		"",
+		"The name of your ECS task definition to modify.",
+	)
+	Cmd.PersistentFlags().BoolVar(
+		&dryRunFlag,
+		"dry-run",
+		false,
+		"Perform a dry run: show what will be done, but do not modify ECS.",
+	)
 
 	// Support for credentials in a nonstandard location
 	Cmd.PersistentFlags().StringVar(&awsCredentialsFlag, "aws-credentials", "", "Location of AWS credentials file.")
@@ -83,18 +91,9 @@ func init() {
 
 func addAgentToECS(cmd *cobra.Command, args []string) error {
 	// Check for API key
-	key, secret, err := cmderr.RequireAPICredentials("The Akita agent must have an API key in order to capture traces.")
+	_, _, err := cmderr.RequireAPICredentials("The Akita agent must have an API key in order to capture traces.")
 	if err != nil {
 		return err
-	}
-	if key == "" || secret == "" {
-		printer.Errorf("No Akita API key configured. The Akita agent must have an API key in order to capture traces.\n")
-		if env.InDocker() {
-			printer.Infof("Please set the AKITA_API_KEY_ID and AKITA_API_KEY_SECRET environment variables on the Docker command line.\n")
-		} else {
-			printer.Infof("Use the AKITA_API_KEY_ID and AKITA_API_KEY_SECRET environment variables, or run 'akita login'.\n")
-		}
-		return cmderr.AkitaErr{Err: errors.New("Could not find an Akita API key to use.")}
 	}
 
 	// Check project's existence
@@ -106,9 +105,20 @@ func addAgentToECS(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		// TODO: we _could_ offer to create it, instead.
 		if strings.Contains(err.Error(), "cannot determine project ID") {
-			return cmderr.AkitaErr{Err: fmt.Errorf("Could not find the project %q in the Akita cloud. Please create it from the Akita web console before proceeding.", projectFlag)}
+			return cmderr.AkitaErr{
+				Err: fmt.Errorf(
+					"Could not find the project %q in the Akita cloud. Please create it from the Akita web console before proceeding.",
+					projectFlag,
+				),
+			}
 		} else {
-			return cmderr.AkitaErr{Err: errors.Wrapf(err, "Could not look up the project %q in the Akita cloud", projectFlag)}
+			return cmderr.AkitaErr{
+				Err: errors.Wrapf(
+					err,
+					"Could not look up the project %q in the Akita cloud",
+					projectFlag,
+				),
+			}
 		}
 	}
 
--- a/cmd/internal/ecs/ecs.go
+++ b/cmd/internal/ecs/ecs.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/akitasoftware/akita-cli/cfg"
 	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
 	"github.com/akitasoftware/akita-cli/env"
 	"github.com/akitasoftware/akita-cli/printer"
@@ -84,7 +83,10 @@ func init() {
 
 func addAgentToECS(cmd *cobra.Command, args []string) error {
 	// Check for API key
-	key, secret := cfg.GetAPIKeyAndSecret()
+	key, secret, err := cmderr.RequireAPICredentials("The Akita agent must have an API key in order to capture traces.")
+	if err != nil {
+		return err
+	}
 	if key == "" || secret == "" {
 		printer.Errorf("No Akita API key configured. The Akita agent must have an API key in order to capture traces.\n")
 		if env.InDocker() {
@@ -100,7 +102,7 @@ func addAgentToECS(cmd *cobra.Command, args []string) error {
 		return errors.New("Must specify the name of your Akita project with the --project flag.")
 	}
 	frontClient := rest.NewFrontClient(rest.Domain, telemetry.GetClientID())
-	_, err := util.GetServiceIDByName(frontClient, projectFlag)
+	_, err = util.GetServiceIDByName(frontClient, projectFlag)
 	if err != nil {
 		// TODO: we _could_ offer to create it, instead.
 		if strings.Contains(err.Error(), "cannot determine project ID") {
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: cmd/internal/ecs/ecs.go [hunk 1]
```diff
@@ -84,7 +83,10 @@ func init() {
...
 func addAgentToECS(cmd *cobra.Command, args []string) error {
 	// Check for API key
-	key, secret := cfg.GetAPIKeyAndSecret()
+	key, secret, err := cmderr.RequireAPICredentials("The Akita agent must have an API key in order to capture traces.")
+	if err != nil {
+		return err
+	}
 	if key == "" || secret == "" {
 		printer.Errorf("No Akita API key configured. The Akita agent must have an API key in order to capture traces.\n")
 		if env.InDocker() {
```
mgritter, line 92: Looks like you didn't remove the old check for empty key or secret.

cmd/internal/kube/kube.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var Cmd = &cobra.Command{
+	Use:   "kube",
+	Short: "Gateway to Kubernetes related utilities",
```
mgritter, line 11:
```suggestion
	Short: "Install Akita in your Kubernetes cluster",
```

Let's try to make this action-oriented (why would they run this) rather than generic -- that's the pattern we used with 'akita ecs' and 'akita nginx'.

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+var (
+	output    string
+	namespace string
+	// Store a parsed representation of /template/akita-secret.tmpl
+	secretTemplate *template.Template
+)
+
+var secretCmd = &cobra.Command{
+	Use:   "secret",
+	Short: "Generate a Kubernetes secret config for Akita",
```
mgritter, line 26: I'm not sure 'Kubernetes secret config' is clear?

```suggestion
	Short: "Generate a Kubernetes secret containing the Akita credentials",
```

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	return outputFile, nil
+}
+
+func init() {
+	var err error
+
+	secretTemplate, err = template.ParseFS(templateFS, "template/akita-secret.tmpl")
+	if err != nil {
+		log.Fatalf("unable to parse kube secret template: %v", err)
+	}
```
mgritter, line 113: I'm a little worried that putting this in the init() function means that somebody who is not trying to use Kubernetes at all could get a failure here. I think there is not a lot of downside to doing it in the execution of the command instead.

cmd/internal/kube/template/akita-secret.tmpl
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: akita-secrets
+  namespace: {{.Namespace}}
```
mgritter, line 5: This should be optional; it need not be specified for the default namespace.

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+
+	secretCmd.Flags().StringVarP(
+		&namespace,
+		"namespace",
+		"n",
+		"",
+		"The Kuberenetes namespace the secret should be applied to",
+	)
+	_ = secretCmd.MarkFlagRequired("namespace")
```
mgritter, line 122: The default namespace should be permitted, I think.

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/akitasoftware/akita-cli/printer"
+
+	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	output    string
+	namespace string
```
mgritter, line 19: Our convention for flag variables seems to be outputFlag, namespaceFlag

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	outputFilePath := filepath.Join(absOutputDir, outputName)
+	if _, statErr := os.Stat(outputFilePath); statErr == nil {
+		return nil, errors.Errorf("output file %s already exists", outputFilePath)
+	}
+
+	// Create the output directory if it doesn't exist
+	err = os.MkdirAll(absOutputDir, 0755)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create the output directory")
+	}
```
mgritter, line 96: It seems a little unusual to me for a tool of this type to create the directory.  Probably I would just raise an error at this point.

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	secretCmd.Flags().StringVarP(
+		&namespace,
+		"namespace",
+		"n",
+		"",
+		"The Kuberenetes namespace the secret should be applied to",
+	)
+	_ = secretCmd.MarkFlagRequired("namespace")
+
+	secretCmd.Flags().StringVarP(&output, "output", "o", "akita-secret.yml", "File to output the generated secret.")
```
mgritter, line 124: This should have come up in the design review -- sorry.  I think the correct default, which is most idiomatic to Kubernetes tools, is to print to standard output.

The idiomatic usage we are aiming for is something like "akita kube secret | kubectl apply -f -"

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, errors.Wrap(err, "failed to create the output file")
+	}
+
+	return outputFile, nil
+}
+
+func init() {
+	var err error
+
+	secretTemplate, err = template.ParseFS(templateFS, "template/akita-secret.tmpl")
```
mgritter, line 110: I'm mostly OK with using a template here.  But, I think this will be much harder to pull off for the next command, and I would like the two implementations to be consistent.  If we have to have YAML parsing and output, let's start here with the easy case.  :)

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		log.Fatalf("unable to parse kube secret template: %v", err)
+	}
+
+	secretCmd.Flags().StringVarP(
+		&namespace,
+		"namespace",
+		"n",
+		"",
+		"The Kuberenetes namespace the secret should be applied to",
```
mgritter, line 120:
```suggestion
		"The Kubernetes namespace the secret should be applied to",
```

mgritter: I'm not sure whether "applied to" or "stored in" is more idiomatic.

cmd/internal/kube/kube.go
```diff
@@ -0,0 +1,15 @@
...
+	RunE: func(_ *cobra.Command, _ []string) error {
+		return cmderr.AkitaErr{Err: errors.New("no subcommand specified")}
+	},
```
mgritter, line 14: Maybe specify 'k8s' as an alias, or 'kubernetes'?

mgritter (changes requested):
Looks pretty good.

The must-fix items: let's add the ability to write to stdout, and use the default namespace.  (And the redundant code in addAgentToECS.)

I would like to see us tackle YAML output with a library here on an easy case before we have to tackle parsing and rewriting a Deployment resource. Unless you were planning on using just string manipulation? But I don't insist on it.

cmd/internal/kube/kube.go
versilis: I've added aliases for both in [935b4bb](https://github.com/akitasoftware/akita-cli/pull/202/commits/935b4bbcb6c6bb93fa562329ad22bce93911b09d)

cmd/internal/kube/template/akita-secret.tmpl
versilis: I've made the namespace flag optional in [6596d5e](https://github.com/akitasoftware/akita-cli/pull/202/commits/6596d5e87dd4c624200a42f5bc4ef2a313b18fce)

cmd/internal/kube/secret.go
versilis: Currently, there's a discrepancy between the Kubernetes openapi spec and the Go packages API model representation. Here's a link to the issue that covers it: https://github.com/kubernetes/kubernetes/issues/109427

I've opened a separate PR to address using the Kubernetes API for generating secrets: https://github.com/akitasoftware/akita-cli/pull/204

cmd/internal/kube/secret.go
versilis: I've updated these flag names in: [69857a4](https://github.com/akitasoftware/akita-cli/pull/202/commits/69857a46fa69b814319c8ebe9ca018b97445e366)

cmd/internal/ecs/ecs.go
versilis: I've removed the old check in: [8d66125](https://github.com/akitasoftware/akita-cli/pull/202/commits/8d661256aba02387c10e5efe17e37b9897feaada)

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,148 @@
...
+			return err
+		}
+
+		output, err := handleSecretGeneration(namespaceFlag, key, secret, outputFlag)
+		if err != nil {
+			return err
+		}
+
+		// Output the generated secret to the console
+		printer.RawOutput(output)
```
mgritter, line 40: Sorry, I think I was unclear.  The two uses cases are:

1. Apply directly 
akita kube secret | kubectl apply -f -

2. Apply via a file (convenience, they could always pipe to a file.)
akita kube secret -f mysecret.yaml
kubectl apply -f mysecret.yaml

In case #2 we should not print to standard output as well, it should be one or the other.  We can support this in a few different ways, I don't much care whether (a) standard out is the default if `-f` not specified, or (b) `-f -` writes to standard output..

I kind of think like in case #1 we should not write the file that was not asked for as well.

Come talk to me if what should happen in these two cases are still unclear.

mgritter (changes requested):

cmd/internal/kube/secret.go
```diff
@@ -0,0 +1,148 @@
...
+	_, err = secretFile.Write(buf.Bytes())
+	if err != nil {
+		return "", cmderr.AkitaErr{Err: errors.Wrap(err, "failed to read generated secret file")}
+	}
+
+	return buf.String(), nil
+}
+
+// Creates a file at the give path to be used for storing of the generated Secret config
+// If any child dicrectories do not exist, it will be created.
```
mgritter, line 105: Is this comment correct?  I think you mean "it will not be created".

pcap/stream_test.go
```diff
@@ -127,6 +128,7 @@ func runTCPFlowTestCase(c tcpFlowTestCase) error {
 }
 
 func TestTCPFlow(t *testing.T) {
+	telemetry.Init(false)
```
mgritter, line 131: I think I understand the need here and in pase_http_test since the error functions in parsing send telemetry.

Is there a fix to telemetry that checks whether uninitialized instead?  If it's not easy to do, we can keep this, it just seems a bit odd to be initializing telemetry in a situation where we don't really want it.

mgritter (approved):
Looks good!

PR #207, opened 2023-03-23
Add utilities for Injecting Kubernetes Deployments

versilis:
This PR adds utilities for injecting Kubernetes deployments to be used with https://github.com/akitasoftware/akita-cli/pull/206. 

The main component is the `Injector` interface which provides the functionality to traverse YAML files (including those with multiple resources using the `---` directive), and inject sidecar containers into any found Deployments.


cmd/internal/kube/test_resource/resources.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
```
versilis, line 1: I haven't written any unit tests to go alongside this, but it works well for interacting with the CLI

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	kyamlutil "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/kube-openapi/pkg/util/sets"
+	"os"
+	"path/filepath"
+)
+
+type (
+	Injector interface {
```
mgritter, line 20: The interface name was a little confusing to me because this is a bit more like a wrapper around the data structure than like a visitor or actor.  It's more like a "Resources" or "Declarations".  But we can leave it as-is.

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+	multidocReader := kyamlutil.NewYAMLReader(bufio.NewReader(bytes.NewReader(yamlContent)))
+
+	var objList []*unstructured.Unstructured
+	for {
+		raw, err := multidocReader.Read()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, err
```
mgritter, line 54: Is the context for all three of these possible errors clear, or should they be wrapped?

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		objList = append(objList, &unstructured.Unstructured{Object: result})
+	}
+
+	return &injectorImpl{objects: objList}, nil
+}
+
+func (i *injectorImpl) InjectableNamespaces() ([]string, error) {
+	set := make(map[string]struct{})
```
mgritter, line 74: You should probably be using akita's set.Set implementation here.  I'm not sure what facilities kube-openapi's sets provides?  But we certainly have a Set-> slice operation.

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		containers := injectedDeployment.Spec.Template.Spec.Containers
+		injectedDeployment.Spec.Template.Spec.Containers = append(containers, sidecar)
+
+		if err != nil {
+			return nil, err
+		}
```
mgritter, line 123: This looks like a duplicate of the check already done in lines 114-116

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	return sets.StringKeySet(set).List(), nil
+}
+
+func (i *injectorImpl) Inject(sidecar v1.Container) ([]*unstructured.Unstructured, error) {
+	onMap := func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+		out := obj.DeepCopy()
+		gvk := out.GetObjectKind().GroupVersionKind()
```
mgritter, line 101: I don't see a need for a DeepCopy here at all; if it's necessary, can you explain it?  It seems fine for the input to be copy-on-write, unless we are making multiple changes where amortizing the copy makes sense.

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+		var deployment *appsv1.Deployment
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(out.Object, &deployment)
+		if err != nil {
+			return nil, err
+		}
+
+		injectedDeployment := deployment.DeepCopy()
+		if err != nil {
+			return nil, err
+		}
```
mgritter, line 116: I think this copy is even less apropos -- presumably the object was just allocated in FromUnstructured, so this function is the only one with access to it?

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+	if _, staterr := os.Stat(absPath); os.IsNotExist(staterr) {
+		return nil, errors.Wrapf(staterr, "file %s does not exist", absPath)
+	}
+
+	return os.ReadFile(absPath)
+}
+
+// mapUnstructured applies the given transformer function to each unstructured Kubernetes item in the given list.
+// If the transformer function returns an error, the error is returned immediately.
+func mapUnstructured(
```
mgritter, line 186: I like this utility function, but I think it could be replaces by akitasoftware/go-utils slices.MapWithErr?

https://github.com/akitasoftware/go-utils/blob/6e85caefbf2ff699b1923eac1807e14cbdc894e1/slices/map.go#L13

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+		gvk := out.GetObjectKind().GroupVersionKind()
+
+		if !isInjectable(gvk) {
+			return out, nil
+		}
+
+		var deployment *appsv1.Deployment
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(out.Object, &deployment)
+		if err != nil {
+			return nil, err
```
mgritter, line 110: Same question about errors as above, if you see one of them in a customer's error report, will you be able to figure out which of these paths it was?  I would probably wrap them.

cmd/internal/kube/injector/injector.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		return out, nil
+	}
+
+	return mapUnstructured(i.objects, onMap)
+}
+
+func isInjectable(kind schema.GroupVersionKind) bool {
+	return kind.Group == "apps" && kind.Version == "v1" && kind.Kind == "Deployment"
```
mgritter, line 137: Would it work to define a schema.GroupVersionKind as a variable and compare with ==?  This is not a big difference (in fact it would be more lines of code) but you could then omit this helper entirely and do == or switch comparison at the point of use.

cmd/internal/kube/injector/injector_test.go
```diff
@@ -0,0 +1,98 @@
...
+		u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&obj)
+		if err != nil {
+			panic(err)
+		}
+
+		return &unstructured.Unstructured{Object: u}
+	}
+
+	appendContainer := func(deployment *appsv1.Deployment, container v1.Container) *appsv1.Deployment {
+		injectedDeployment := deployment.DeepCopy()
```
mgritter, line 24: The DeepCopy is fine (and necessary!) here.

mgritter (approved):
I didn't find anything big, but if you have time there are some changes I would suggest.

The most important one is supportability; is the context for each error clear?

Next-most important is using code from go-utils.

Final set is avoiding unnecessary deep copies.

cmd/internal/kube/injector/injector.go
versilis: I kind of like the idea of keeping the helper function since it's already being used by two methods (`InjectableNamespaces` and `Inject`). However, I have simplified the check as you've suggested in https://github.com/akitasoftware/akita-cli/pull/207/commits/12f88e2d4b2404332e9551064422d5be0ab6dfc8

cmd/internal/kube/injector/injector.go
versilis: Agreed! I initially intended for this interface to only have a single `Inject` method that took a set of resources and updated them, but it seems much better suited to be a wrapper and manipulator of Kubernetes resources.

I won't address this now, but if I have time later I'll refactor this interface's name into something more well-suited.

cmd/internal/kube/injector/injector.go
versilis: Updated in [8edde25](https://github.com/akitasoftware/akita-cli/pull/207/commits/8edde25a6cae3fd35be0332dcf1ff13d9c2a7832)

cmd/internal/kube/injector/injector.go
versilis: I've updated the injector to use our existing libraries in [1b26251](https://github.com/akitasoftware/akita-cli/pull/207/commits/1b262511defc26e66c122ce152d99f196bf24e76)

cmd/internal/kube/injector/injector.go
versilis: Agreed. I think this was a relic of a previous iteration I had of this code. I've removed any redundant deep copies in [dff7748](https://github.com/akitasoftware/akita-cli/pull/207/commits/dff77482518524a97a1c6b9aef53b74e2845f8ef)

cmd/internal/kube/injector/injector.go
versilis: I've removed this duplicate along with the changes in https://github.com/akitasoftware/akita-cli/pull/207/commits/48aa8aa517b2ec3c77e354106cb1a9131d46e678

PR #213, opened 2023-07-27
[OBS-36] Add postman flags support in `apidump`

mudit-postman:
* Add support for 3 Postman flags in `apidump` command, namely `--collection`, `--key `, and `--environment`
* Get the correct serviceID based on the passed Postman collectionID

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -80,7 +80,7 @@ type Args struct {
 
 	// Optional args
 
-	// If both LocalPath and AkitaURI are set, data is feed to both local traces
+	// If both LocalPath and AkitaURI are set, data is teed to both local traces
 	// and backend trace.
 	// If unset, defaults to a random spec name on Akita Cloud.
 	Out location.Location
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -553,6 +553,13 @@ func (a *apidump) Run() error {
 				return errors.Errorf("Cannot automatically rotate sessions when a session name is provided.")
 			}
 		}
+	} else if args.PostmanCollectionID != "" {
+		args.Out.AkitaURI = &akiuri.URI{
+			ObjectType: akiuri.TRACE.Ptr(),
+			// Placeholder name for postman collectionIds
+			ServiceName: "Postman_" + args.PostmanCollectionID,
+			ObjectName:  util.RandomLearnSessionName(),
+		}
 	}
 
 	// If --dogfood is specified, enable assertions in the buffer-pool code.
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -166,7 +166,7 @@ func newSession(args *Args) *apidump {
 
 // Is the target the Akita backend as expected, or a local HAR file?
 func (a *apidump) TargetIsRemote() bool {
-	return a.Out.AkitaURI != nil
+	return a.Out.AkitaURI != nil || a.PostmanCollectionID != ""
 }
 
 // Lookup the service and create a learn client targeting it.
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -80,10 +80,13 @@ type Args struct {
 
 	// Optional args
 
-	// If both LocalPath and AkitaURI are set, data is teed to both local traces
+	// If both LocalPath and AkitaURI are set, data is feed to both local traces
 	// and backend trace.
 	// If unset, defaults to a random spec name on Akita Cloud.
 	Out location.Location
+
+	// Args used to using agent with Postman
+	PostmanCollectionID string
 
 	Interfaces     []string
 	Filter         string
@@ -172,12 +175,24 @@ func (a *apidump) LookupService() error {
 		return nil
 	}
 	frontClient := rest.NewFrontClient(a.Domain, a.ClientID)
-	backendSvc, err := util.GetServiceIDByName(frontClient, a.Out.AkitaURI.ServiceName)
-	if err != nil {
-		return err
+
+	if a.PostmanCollectionID != "" {
+		backendSvc, err := util.GetServiceIDByPostmanCollectionID(frontClient, a.PostmanCollectionID)
+		if err != nil {
+			return err
+		}
+
+		a.backendSvc = backendSvc
+	} else {
+		backendSvc, err := util.GetServiceIDByName(frontClient, a.Out.AkitaURI.ServiceName)
+		if err != nil {
+			return err
+		}
+
+		a.backendSvc = backendSvc
 	}
-	a.backendSvc = backendSvc
-	a.learnClient = rest.NewLearnClient(a.Domain, a.ClientID, backendSvc)
+
+	a.learnClient = rest.NewLearnClient(a.Domain, a.ClientID, a.backendSvc)
 	return nil
 }
 
--- a/cmd/internal/apidump/apidump.go
+++ b/cmd/internal/apidump/apidump.go
@@ -217,7 +217,7 @@ func init() {
 		&postmanCollectionID,
 		"collection",
 		"",
-		"Your Postman collectionID. Both --collection and --key must be specified.")
+		"Your Postman collectionID.")
 
 	Cmd.MarkFlagsMutuallyExclusive("out", "project", "collection")
 
--- a/cmd/internal/apidump/apidump.go
+++ b/cmd/internal/apidump/apidump.go
@@ -2,6 +2,7 @@ package apidump
 
 import (
 	"os"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/akitasoftware/akita-cli/apidump"
 	"github.com/akitasoftware/akita-cli/apispec"
+	"github.com/akitasoftware/akita-cli/cfg"
 	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
 	"github.com/akitasoftware/akita-cli/cmd/internal/pluginloader"
 	"github.com/akitasoftware/akita-cli/location"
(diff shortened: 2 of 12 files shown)


apidump/apidump.go
```diff
@@ -80,11 +80,14 @@ type Args struct {
 
 	// Optional args
 
-	// If both LocalPath and AkitaURI are set, data is teed to both local traces
+	// If both LocalPath and AkitaURI are set, data is feed to both local traces
```
mgritter, line 83: No, it really did mean "teed" as in the command-line utility "tee"

cmd/internal/apidump/apidump.go
```diff
@@ -80,6 +84,19 @@ var Cmd = &cobra.Command{
 			outFlag.AkitaURI = &uri
 		}
 
+		if postmanAPIKey != "" {
+			cfg.WritePostmanAPIKeyAndEnvironment("default", postmanAPIKey, postmanEnvironment)
+		}
```
mgritter, line 89: I don't think we should write the config file here, unfortunately.  That belongs in a login command if at all.

cmd/internal/apidump/apidump.go
```diff
@@ -201,6 +219,28 @@ func init() {
...
+
+	Cmd.MarkFlagsMutuallyExclusive("out", "project", "collection")
+
+	Cmd.Flags().StringVar(
+		&postmanAPIKey,
+		"key",
+		"",
+		"Your Postman API Key. Both --collection and --key must be specified.")
+
+	Cmd.MarkFlagsRequiredTogether("collection", "key")
```
mgritter, line 236: I think that it would be best if we did not accept sensitive material on the command line. We might have to get Paul to update the command line in the instructions from

`akita kube inject --collection foo --key <key>`

to 

`POSTMAN_API_KEY=<key> akita kube inject --collection foo`

cmd/internal/apidump/apidump.go
```diff
@@ -201,6 +219,28 @@ func init() {
...
+
+	Cmd.Flags().StringVar(
+		&postmanEnvironment,
+		"environment",
+		"",
+		"Your Postman Environment. Default value is PREVIEW")
+
```
mgritter, line 243: I think we should probably make this hidden, as it will be irrelevant to all but Postman developers.

We should figure out when the cutover is from PREVIEW being the default, to production.  We can do that in a CLI release after 8/14, PREVIEW is probably best now.

cmd/internal/apidump/apidump.go
```diff
@@ -80,6 +84,19 @@ var Cmd = &cobra.Command{
...
+		}
+
+		// ToDo: Do we need to change akiuri package also to check proper collectionID format??
+		if postmanCollectionID != "" {
+			uri, err := akiuri.Parse(akiuri.Scheme + postmanCollectionID)
+			if err != nil {
+				return errors.Wrap(err, "bad postman collection-id name")
+			}
+			outFlag.AkitaURI = &uri
+		}
```
mgritter, line 98: No, we should further deprecate the use of the Akita URI here.  You're doing the correct thing by passing it as a separate argument to apidump.Args.  We should not need it to also be present in AkitaURI.

cmd/internal/apidump/apidump.go
mgritter: Another option is to make PREVIEW the default is domain is staging.akita.software and use PRODUCTION otherwise.

cmd/internal/apidump/apidump.go
mudit-postman: I had done the changes based on the UI Mockup shared by Paul, [here](https://postman.slack.com/archives/C05EHCJ6KD5/p1689987756340769).
If we need to use `akita login` command then we should do the required changes in UI side also

cmd/internal/apidump/apidump.go
mudit-postman: Since this environment flag is only for running in lower envs not in production, I have decided to pass the environment header only if this argument is provided, otherwise, it will default to PREVIEW in backend ([ref](https://github.com/akitasoftware/superstar/blob/master/services/front/app/auth.py#L164)).

We can make the default postman env value config driven in BE and set the value of the config accordingly.

mudit-postman: We can remove `Default value is PREVIEW` message to remove confusion.

cmd/internal/apidump/apidump.go
mudit-postman: What advantage will we be having by accepting key as env var? In both cases user have to specify the key in command line for `kube inject`.

Why can't we ask user to run `akita login` first, just like we are doing earlier?

cmd/internal/apidump/apidump.go
mudit-postman: `--domain` flag is still something we need to resolve

cmd/internal/apidump/apidump.go
mgritter: The point is that executing 'apidump' or 'inject' should not alter the user's configuration file.  I don't think we need to support 'login' -- though that would be fine if we did.  But the user should be able to provide an alternative ID in an environment variable without having that value persisted.

cmd/internal/apidump/apidump.go
mgritter: The advantage here is that the actual command arguments are visible in contexts where the environment variables (even if specified on the shell) are not -- it's a security best practice.  There is less potential for exposing sensitive information.

(We could go back and rewrite the command line on initialization; some tools do that.)

Our goal here is to have a single command that they can copy from the UI and run.  Adding a second step, particularly one that's interactive like `login`, interferes with that.

cmd/internal/apidump/apidump.go
```diff
@@ -80,6 +84,17 @@ var Cmd = &cobra.Command{
...
 
+		// If --collection was given, set rest.domain and environment config (if given)
+		if postmanCollectionID != "" {
+			if postmanEnvironment != "" {
+				env := strings.ToUpper(postmanEnvironment)
+
+				cfg.WritePostmanEnvironment("default", env)
+			}
+
+			rest.Domain = "staging.akita.software"
```
mgritter, line 95: See my comment in https://github.com/akitasoftware/akita-cli/pull/214, I think to do this right we need to pull the environment -> domain logic up into the root command.

cmd/internal/apidump/apidump.go
```diff
@@ -80,6 +84,17 @@ var Cmd = &cobra.Command{
 			outFlag.AkitaURI = &uri
 		}
 
+		// If --collection was given, set rest.domain and environment config (if given)
+		if postmanCollectionID != "" {
+			if postmanEnvironment != "" {
+				env := strings.ToUpper(postmanEnvironment)
+
+				cfg.WritePostmanEnvironment("default", env)
```
mgritter, line 92: I don't understand why you are writing to the config file here.  `apidump` should not alter any configuration files, only `login` should do so.

rest/base_client.go
```diff
@@ -64,13 +65,13 @@ type BaseClient struct {
...
-	// If rawHost is IP, IP:port, localhost, or localhost:port, use that
-	// directly. This is mostly to support tests.
+	// If rawHost is either of these: IP, IP:port, localhost, localhost:port or postman host
+	// use that directly. This is to support postman agent and tests.
 	if h, _, err := net.SplitHostPort(rawHost); err == nil {
 		if h == "localhost" || net.ParseIP(h) != nil {
 			host = rawHost
 		}
-	} else if rawHost == "localhost" || net.ParseIP(rawHost) != nil {
+	} else if rawHost == "localhost" || net.ParseIP(rawHost) != nil || strings.Contains(rawHost, "postman") {
```
mgritter, line 74: This is a good idea. I worry that it's a bit too broad.

Maybe what we can do is lift the conversion from 

domain akita.software => host api.akita.software
domain staging.akita.software => host api.staging.akita.software
domain localhost => host localhost

to a higher level where we're simultaneously doing the environment to domain mapping?  We may just be able to scrap the "prefix api" logic here, since it's not really used for anything but akita-staging and akita-prod.  (Originally the thinking was that customers would have their own isolated clusters and so you'd have something like api.customername.akita.software.)

mgritter (changes requested):
I would have preferred if we had a testable version of apidump even if the domain had to be specified explicitly, rather than trying to do everything at once.

The must-change item here is that apidump should not write a change to config file.

Can we land a change that permits end-to-end testing to start, even if the --domain flag is necessary until our next milestone?

rest/base_client.go
mgritter: That is, the idea is that if the user specifies "staging.akita.software" as the domain then we translate it to the endpoint "api.staging.akita.software" and the same for the special value "akita.software", but have those be the two exceptional cases rather than the current list of exceptions (localhost, IP address, and now "postman" as a substring.)

The logic can treat the two previous cases as the exceptions rather than the ones we want to support going forward.

cmd/internal/apidump/apidump.go
mudit-postman: Wasn't aware that config files should only be updated via `login` command, removed this from both `apidump` and `kube inject`

rest/base_client.go
mudit-postman: Will pick this up in the next milestone, for now, will only target to make things working for demo.

mudit-postman:
Have removed the configuration-related changes from `apidump` and also removed the `environment` flag as discussed earlier.

Regarding the `async_report` endpoint failing it is happening from BE side, same is happening when trying to hit the request via postman. Postman's ACS service is giving error, will continue checking on this.

apidump/apidump.go
```diff
@@ -538,6 +553,13 @@ func (a *apidump) Run() error {
 				return errors.Errorf("Cannot automatically rotate sessions when a session name is provided.")
 			}
 		}
+	} else if args.PostmanCollectionID != "" {
+		args.Out.AkitaURI = &akiuri.URI{
+			ObjectType: akiuri.TRACE.Ptr(),
+			// Placeholder name for postman collectionIds
+			ServiceName: "Postman_" + args.PostmanCollectionID,
```
mgritter, line 560: Do we have the real service name we could use?

mgritter (approved):

apidump/apidump.go
mudit-postman: We can get it from the lookup service response. Will raise a new PR for this change

PR #214, opened 2023-07-27
[OBS-37|38] Add support for Postman flags in `kube inject` command

mudit-postman:
* Add support for 3 Postman flags in `kubectl inject` command, namely --collection, --key , and --environment
* Setting postman API key as an k8s secret
* Passing other two params normally in args
* Added env var support for postman API key
* Other refactoring and typos fix wherever necessary

### DIFF
--- a/cmd/internal/kube/inject.go
+++ b/cmd/internal/kube/inject.go
@@ -3,7 +3,6 @@ package kube
 import (
 	"bytes"
 
-	"github.com/akitasoftware/akita-cli/cfg"
 	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
 	"github.com/akitasoftware/akita-cli/cmd/internal/kube/injector"
 	"github.com/akitasoftware/akita-cli/printer"
@@ -32,7 +31,6 @@ var (
 
 	// Postman related flags
 	postmanCollectionID string
-	postmanAPIKey       string
 	postmanEnvironment  string
 )
 
@@ -45,11 +43,6 @@ var injectCmd = &cobra.Command{
 			return cmderr.AkitaErr{
 				Err: errors.New("exactly one of --project or --collection must be specified"),
 			}
-		}
-
-		// Adding postman API key and Environment in config for further uses
-		if postmanCollectionID != "" {
-			cfg.WritePostmanAPIKeyAndEnvironment("default", postmanAPIKey, postmanEnvironment)
 		}
 
 		secretOpts := resolveSecretGenerationOptions(secretInjectFlag)
@@ -85,8 +78,13 @@ var injectCmd = &cobra.Command{
 			}
 
 			if postmanCollectionID != "" {
+				key, err := cmderr.RequirePostmanAPICredentials("Postman API credentials are required to generate secret.")
+				if err != nil {
+					return err
+				}
+
 				for _, namespace := range namespaces {
-					r, err := handlePostmanSecretGeneration(namespace, postmanAPIKey)
+					r, err := handlePostmanSecretGeneration(namespace, key)
 					if err != nil {
 						return err
 					}
@@ -95,7 +93,7 @@ var injectCmd = &cobra.Command{
 					secretBuf.Write(r)
 				}
 			} else {
-				key, secret, err := cmderr.RequireAPICredentials("API credentials are required to generate secret.")
+				key, secret, err := cmderr.RequireAkitaAPICredentials("Akita API credentials are required to generate secret.")
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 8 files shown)


cmd/internal/kube/inject.go
```diff
@@ -28,13 +29,29 @@ var (
...
 	RunE: func(_ *cobra.Command, args []string) error {
+		if projectNameFlag == "" && postmanCollectionID == "" {
+			return cmderr.AkitaErr{
+				Err: errors.New("exactly one of --project or --collection must be specified"),
+			}
+		}
+
+		// Adding postman API key and Environment in config for further uses
+		if postmanCollectionID != "" {
+			cfg.WritePostmanAPIKeyAndEnvironment("default", postmanAPIKey, postmanEnvironment)
```
mgritter, line 52: I think we need to avoid writing the config file here too.

cmd/internal/kube/inject.go
```diff
@@ -62,24 +79,36 @@ var injectCmd = &cobra.Command{
...
 			namespaces, err := injectr.InjectableNamespaces()
 			if err != nil {
 				return err
 			}
 
-			for _, namespace := range namespaces {
-				r, err := handleSecretGeneration(namespace, key, secret)
+			if postmanCollectionID != "" {
+				for _, namespace := range namespaces {
+					r, err := handlePostmanSecretGeneration(namespace, postmanAPIKey)
```
mgritter, line 89: I don't see any check that the postmanAPIKey is set (like there is for the Akita credentials below?)

cmd/internal/kube/template/postman-secret.tmpl
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: akita-postman-secrets
```
mgritter, line 4: I know that I wrote 'akita-secrets' in the doc.  But looking at this, I think we should get the branding right here and call it something like postman-live-collection-agent-secrets, or otherwise avoid the 'Akita' name.

cmd/internal/kube/inject.go
```diff
@@ -184,6 +221,43 @@ func createSidecar(projectName string) v1.Container {
...
+				Exec: &v1.ExecAction{
+					Command: []string{
+						"/bin/sh",
+						"-c",
+						"AKITA_PID=$(pgrep akita) && kill -2 $AKITA_PID && tail -f /proc/$AKITA_PID/fd/1",
+					},
+				},
+			},
+		},
+		Args: []string{"apidump", "--collection", postmanCollectionID, "--environment", postmanEnvironment},
```
mgritter, line 252: As things stand currently in the akita-cli side we also need to specify `--domain staging.akita.software` so that the CLI talks to the correct API endpoint.

I do not like this answer, so the alternative is to derive the domain (staging.akita.software) from the choice of postman environment in akita-cli.

cmd/internal/kube/inject.go
mudit-postman: Since we will be getting it as the arguments itself, that's why I had not checked it here.
But since now we will be removing the Postman API key as argument, so will de the required changes.

cmd/internal/kube/inject.go
mudit-postman: Currently, I am trying to read both Postman API Key and Environment from the config file, is it okay? Or, should we keep only the API key in the config file?
Also I am thinking to only write postman environment to config file when it is sent by user, cause anyhow in production it will never be used.

cmd/internal/kube/inject.go
```diff
@@ -28,13 +32,41 @@ var (
...
+
+		// If --collection was given, set rest.domain and environment config (if given)
+		if postmanCollectionID != "" {
+			if postmanEnvironment != "" {
+				env := strings.ToUpper(postmanEnvironment)
+
+				cfg.WritePostmanEnvironment("default", env)
+			}
+
+			rest.Domain = "staging.akita.software"
```
mgritter, line 60: I am a bit worried about this.

What I would like is to change the _default_ domain based on the postman environment. What you've done here will overwrite the user's --domain flag even if they gave it explicitly.

I don't think this is something that the flags library makes easy.  We probably have to use "" as the default domain for the flag, so we can tell whether it's been set or not.  (Unless there's a way to query the flag?)

```
        rootCmd.PersistentFlags().StringVar(&rest.Domain, "domain", defaultDomain, "Your assigned Akita domain (e.g. company.akita.software)")
        rootCmd.PersistentFlags().MarkHidden("domain")
```

We probably also need to make the environment flag a root command flag.  Then in `preRun` we can examine the environment flag and the rest domain flag and set rest.Domain appropriately.  This needs to happen before the call to telemetry.Init(), which will make an API call.

https://github.com/akitasoftware/akita-cli/blob/890c2fbfe21c506453cfda526c8d3b55085d0e1e/cmd/root.go#L76C7-L77C22

However, I don't think this matters for the `kube` commands which as far as we know don't make API calls at all, except for the telemetry setup?

mgritter: Ah, I see this is not true because you introduced a call to `lookupService` below.

cmd/internal/kube/inject.go
```diff
@@ -28,13 +32,41 @@ var (
...
+		}
+
+		// Lookup service *first* (if we are remote) ensuring that collectionId
+		// or projectName is correct and exists.
+		err := lookupService(postmanCollectionID, projectNameFlag)
+		if err != nil {
+			return err
+		}
```
mgritter, line 68: I don't recall what was behind @versilis 's decision _not_ to make this check.  Perhaps we wanted `kube inject` to work even in environments where the API endpoint was unreachable.  

@versilis do you recall if there was a requirement or use case here?

mgritter (commented):
Looks pretty good, but the introduction of the domain override and the check both raise questions. Let's sync up on these -- the domain default logic probably needs to be hoisted to the root command to work correctly.

cmd/internal/kube/inject.go
mudit-postman: As discussed will be removing the `environment` flag and will pick domain setting in a separate task for next milestone.

cmd/internal/kube/inject.go
mudit-postman: As discussed will be keeping this check now

mgritter (approved):

PR #215, opened 2023-08-03
[OBS-83] Create mirror service from cli

mudit-postman:
In this PR we have added support for the following:
* Ability to create service in Akita BE if the request is coming for postman agent
* Add env var in caching along with `collectionId`
* We are using random name for `serviceName`

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -12,6 +12,7 @@ import (
 	cache "github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
 
+	"github.com/akitasoftware/akita-cli/cfg"
 	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/rest"
 	"github.com/akitasoftware/akita-cli/telemetry"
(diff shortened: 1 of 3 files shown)


util/util.go
```diff
@@ -122,19 +124,27 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
...
 		return result, nil
 	}
-	telemetry.Failure("Unknown postman collectionID")
-	return akid.ServiceID{}, errors.Errorf("cannot determine project ID for Postman collectionID %s", collectionID)
+
+	printer.Debugf("Found no service for given collectionID: %s, creating a new service", collectionID)
+	// Create service for given postman collectionID
+	service, err := c.CreateService(ctx, "Postman-"+randomName(), collectionID, env)
+	if err != nil {
+		return akid.ServiceID{}, errors.Wrap(err, fmt.Sprintf("failed to create or get service for given collectionID: %s", collectionID))
```
mgritter, line 144: What is the end-to-end user experience like when this happens?  Should we look for this error higher up and provide some more guidance, or emit a more detailed message here?

The minimum we owe a user is: what do we want them to do?  Contact Postman?  (ugh... I haven't yet thought about brining up that subject.)  Try again later?

Can we distinguish permission problems from other sources of error?  We should be able to provide a good message when the problem is "you don't have permission to edit that collection".

mgritter (approved):
Looks good.

We probably need some work to smooth the end-to-end experience. We can't depend on a customer parsing the error message and deciding what to do on failure.  Please add a ticket for that as follow-up work, but there might be some changes we can make immediately.

PR #222, opened 2023-08-17
fix: omit empty env value

mudit-postman:
Currently if we don't send any value for env then it is sending an empty string which is throwing error in BE due to wrong enum type.
So need to add `omitempty` in model

### DIFF
--- a/rest/models.go
+++ b/rest/models.go
@@ -7,7 +7,7 @@ import (
 
 type PostmanMetaData struct {
 	CollectionID string `json:"collection_id"`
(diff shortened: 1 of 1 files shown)


mgritter (approved):

mgritter:
We could also have removed Environment from this struct entirely, as it's not something the client ever has to send?

mudit-postman:
> We could also have removed Environment from this struct entirely, as it's not something the client ever has to send.

This model was used to store the response also, where we are getting the environment from BE. So have kept it as it is.

PR #223, opened 2023-08-23
chore: remove env check while looping collnId and serviceId mapping

mudit-postman:
* Since we are moving away from the environment flag, whenever we try to get `postman_env's` value we get the empty string `""`
* This is causing an issue while checking for the mapping between service and collectionId, cause we will be getting environment value also from BE and hence `strings.EqualFold(svcEnv, env)` will never be true and we will create a new service every time cli starts
* So in this PR we have removed this check of env vars from here. We don't really need this check because after 9/15 release  `postman_env` field will be irrelevant (every environment has its own infra) and AFAIK, for the existing stage services there are no collections with the same Id within an org

**PS:** One other option we have is, setting a default value for `postman_env` in `initCreds()` [function](https://github.com/akitasoftware/akita-cli/blob/bugfix/remove-env-check-in-service-list/cfg/credentials.go#L60) if it's value is empty

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -122,11 +122,10 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string) 
 
 		// Normalize collectionID.
 		svcCollectionID := strings.ToLower(svc.PostmanMetaData.CollectionID)
-		svcEnv := strings.ToUpper(svc.PostmanMetaData.Environment)
 
 		if strings.EqualFold(collectionID, svcCollectionID) {
 			result = svc.ID
-			postmanCollectionIDCache.Set(svcEnv+"|"+svcCollectionID, svc.ID, cache.DefaultExpiration)
+			postmanCollectionIDCache.Set(svcCollectionID, svc.ID, cache.DefaultExpiration)
 		}
 	}
 
--- a/util/util.go
+++ b/util/util.go
@@ -12,7 +12,6 @@ import (
 	cache "github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
 
-	"github.com/akitasoftware/akita-cli/cfg"
 	"github.com/akitasoftware/akita-cli/printer"
 	"github.com/akitasoftware/akita-cli/rest"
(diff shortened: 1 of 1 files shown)


util/util.go
```diff
@@ -126,7 +124,7 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
 		svcCollectionID := strings.ToLower(svc.PostmanMetaData.CollectionID)
 		svcEnv := strings.ToUpper(svc.PostmanMetaData.Environment)
 
-		if strings.EqualFold(collectionID, svcCollectionID) && strings.EqualFold(svcEnv, env) {
+		if strings.EqualFold(collectionID, svcCollectionID) {
 			result = svc.ID
 			postmanCollectionIDCache.Set(svcEnv+"|"+svcCollectionID, svc.ID, cache.DefaultExpiration)
```
mgritter, line 129: It seems like the cache key should be consistent with what we're checking.

That looks like it was an existing bug, but let's see if we can fix it up.

util/util.go
```diff
@@ -126,7 +124,7 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
 		svcCollectionID := strings.ToLower(svc.PostmanMetaData.CollectionID)
 		svcEnv := strings.ToUpper(svc.PostmanMetaData.Environment)
```
mgritter, line 125: Remove this if we're not going to check it? =

mgritter (changes requested):
Hm... it seems like using "" as the default rather than building in "PREVIEW" for this release was a mistake, given that "" is not a valid environment name.

I am OK with the working assumption that collection ID's won't collide, even if they could in different environments, and as you point out production users will only connect to the production environment.  (But the CLI must still support connecting to any one of the environments for development and testing.)

Another alternative is to use the environment as a filter on the return value iff it is nonempty, but this is fine.

util/util.go
mudit-postman: Agree, this is a bug. Removed env from the cache also

mgritter (approved):

PR #226, opened 2023-09-11
Bubble up permission error to the user while creating a new service

shreys7:
image

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -139,7 +139,16 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string) 
 	// Create service for given postman collectionID
 	resp, err := c.CreateService(ctx, name, collectionID)
 	if err != nil {
-		return akid.ServiceID{}, errors.Wrap(err, fmt.Sprintf("failed to create or get service for given collectionID: %s", collectionID))
+		printer.Stderr.Debugf("Failed to create service for given collectionID: %s\n", err)
+
+		if httpErr, ok := err.(rest.HTTPError); ok && httpErr.StatusCode == 403 {
+			errMsg := fmt.Sprintf("You cannot send traffic to the collection with ID %s."+
+				" Ensure that your collection ID is correct and that you have edit permissions on the collection."+
+				" If you do not have edit permissions, please contact the workspace administrator to add you as a collection editor.", collectionID)
+			return akid.ServiceID{}, errors.New(errMsg)
+		}
+
+		return akid.ServiceID{}, errors.Wrap(err, fmt.Sprintf("failed to create service for given collectionID: %s", collectionID))
 	}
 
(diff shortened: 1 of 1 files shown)


util/util.go
```diff
@@ -139,7 +139,16 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
...
+		printer.Stderr.Debugf("Failed to create service for given collectionID: %s\n", err)
+
+		if httpErr, ok := err.(rest.HTTPError); ok && httpErr.StatusCode == 403 {
+			errMsg := fmt.Sprintf("You cannot send traffic to the collection with ID %s."+
+				" Ensure that your collection ID is correct and that you have edit permissions on the collection."+
+				" If you do not have edit permissions, please contact the workspace administrator to add you as a collection editor.", collectionID)
+			return akid.ServiceID{}, errors.New(errMsg)
+		}
+
+		return akid.ServiceID{}, errors.Wrap(err, fmt.Sprintf("failed to create service for given collectionID: %s", collectionID))
```
mgritter, line 151: Should be able to use errors.Wrapf() here instead.

Please update the text here to give more context as well.  A Postman user doesn't know what a "service" is, or what they should do about this feailure.  For example, we could mention that this was an unexpected error, and that they should contact our support address.  Or explain that there was a setup failure.

util/util.go
```diff
@@ -139,7 +139,16 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
 	// Create service for given postman collectionID
 	resp, err := c.CreateService(ctx, name, collectionID)
 	if err != nil {
-		return akid.ServiceID{}, errors.Wrap(err, fmt.Sprintf("failed to create or get service for given collectionID: %s", collectionID))
+		printer.Stderr.Debugf("Failed to create service for given collectionID: %s\n", err)
+
+		if httpErr, ok := err.(rest.HTTPError); ok && httpErr.StatusCode == 403 {
+			errMsg := fmt.Sprintf("You cannot send traffic to the collection with ID %s."+
```
mgritter, line 145: Use fmt.Errorf() instead?

mgritter (changes requested):

util/util.go
```diff
@@ -96,6 +96,8 @@ func GetServiceIDByName(c rest.FrontClient, name string) (akid.ServiceID, error)
 func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string) (akid.ServiceID, error) {
 	// Normalize the collectionID.
 	collectionID = strings.ToLower(collectionID)
+	unexpectedErrMsg := "Something went wrong while starting the Agent. " +
+		"Please contact postman support(observability-support@postman.com) with the error details"
```
mgritter, line 100:
```suggestion
		"Please contact Postman support (observability-support@postman.com) with the error details"
```

mgritter (approved):

PR #229, opened 2023-09-13
[OBS-270] Use Postman LCA image locations

liujed:


cmd/internal/ecs/add.go
```diff
@@ -531,7 +530,7 @@ func getTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowStat
 	// Check that the Akita CLI is not already present
 	for _, container := range output.ContainerDefinitions {
 		image := aws.ToString(container.Image)
-		if matchesImage(image, akitaECRImage) || matchesImage(image, akitaDockerImage) {
+		if matchesImage(image, akitaECRImage) {
```
mgritter, line 533: If I believed we had any existing users who were going to change, this would be a little concerning -- what we ought to do for robustness is check both old _and_ new names to see whether the agent is installed in the task definition.  But, the ECS onboarding flow did not get a lot of uptake.

mgritter (approved):

cmd/internal/ecs/add.go
liujed: In 84300a7. Re-requesting review to make sure I understood what you meant.

mgritter (approved):
Yes, that will not re-install the agent if it is already installed.  (If EC2 is our next target then we should consider whether upgrading akita->postman is a use case we want to take on.)

PR #232, opened 2023-09-14
[OBS-137] Handle 409 response of create service to again check for service

mudit-postman:
* Create a new model to parse error response
* Handle the 409 conflict case. Check for error message and call get services again 
* Try only one time after that bail out with error
* Other small fixes: Add to cache after create service flow, small model change

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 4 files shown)


util/util.go
```diff
@@ -128,24 +115,72 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string)
...
+
+		var errorResponse rest.CreateServiceErrorResponse
+		if err := json.Unmarshal(httpErr.Body, &errorResponse); err != nil {
+			printer.Stderr.Debugf(failedToCreateServiceErrMsg, err)
+			return akid.ServiceID{}, errors.Wrap(err, unexpectedErrMsg)
+		}
 
-		if httpErr, ok := err.(rest.HTTPError); ok && httpErr.StatusCode == 403 {
-			error := fmt.Errorf("You cannot send traffic to the collection with ID %s. "+
+		if httpErr.StatusCode == 409 && errorResponse.Message == "collection_already_mapped" {
```
mgritter, line 169: Please add a comment explaining the use case which requires this (a race between multiple agents trying to create the mirrored service.)

mgritter (approved):

mgritter:
Merging so we can build a new agent release this morning

PR #234, opened 2023-09-15
Add collection flag for postman ECS agent

gmann42:
- Adds collection flag

### DIFF
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -99,7 +99,7 @@ const (
 	// Postman Live Collections Agent image locations
 	akitaECRImage    = "public.ecr.aws/akitasoftware/akita-cli"
 	akitaDockerImage = "akitasoftware/cli"
-	postmanECRImage = "docker.postman.com/postman-lc-agent"
+	postmanECRImage  = "docker.postman.com/postman-lc-agent"
 )
 
 // Run the "add to ECS" workflow until we complete or get an error.
@@ -841,20 +841,24 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
 	})
 
 	pKey, pEnv := cfg.GetPostmanAPIKeyAndEnvironment()
+	envs := []types.KeyValuePair{}
+	if pEnv != "" {
+		envs = append(envs, []types.KeyValuePair{
+			{Name: aws.String("POSTMAN_ENV"), Value: &pEnv},
+		}...)
+	}
 	input.ContainerDefinitions = append(input.ContainerDefinitions, types.ContainerDefinition{
 		Name: aws.String("postman-lc-agent"),
 		// TODO: Cpu and Memory should be omitted for Fargate; they take their default values for EC2 if omitted.
 		// For now we can leave the defaults in place, but they might be a bit large for EC2.
 		EntryPoint: []string{"/postman-lc-agent", "apidump", "--collection", collectionId},
-		Environment: []types.KeyValuePair{
+		Environment: append(envs, []types.KeyValuePair{
 			{Name: aws.String("POSTMAN_API_KEY"), Value: &pKey},
-			{Name: aws.String("POSTMAN_ENV"), Value: &pEnv},
-
 			// Setting these environment variables will cause the traces to be tagged.
-			{Name: aws.String("AKITA_AWS_REGION"), Value: &wf.awsRegion},
-			{Name: aws.String("AKITA_ECS_SERVICE"), Value: &wf.ecsService},
-			{Name: aws.String("AKITA_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
-		},
+			{Name: aws.String("POSTMAN_AWS_REGION"), Value: &wf.awsRegion},
+			{Name: aws.String("POSTMAN_ECS_SERVICE"), Value: &wf.ecsService},
+			{Name: aws.String("POSTMAN_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
+		}...),
 		Essential: aws.Bool(false),
 		Image:     aws.String(postmanECRImage),
 	})
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -855,9 +855,9 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
 		Environment: append(envs, []types.KeyValuePair{
 			{Name: aws.String("POSTMAN_API_KEY"), Value: &pKey},
 			// Setting these environment variables will cause the traces to be tagged.
-			{Name: aws.String("POSTMAN_AWS_REGION"), Value: &wf.awsRegion},
-			{Name: aws.String("POSTMAN_ECS_SERVICE"), Value: &wf.ecsService},
-			{Name: aws.String("POSTMAN_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
+			{Name: aws.String("AKITA_AWS_REGION"), Value: &wf.awsRegion},
+			{Name: aws.String("AKITA_ECS_SERVICE"), Value: &wf.ecsService},
+			{Name: aws.String("AKITA_ECS_TASK"), Value: &wf.ecsTaskDefinitionFamily},
 		}...),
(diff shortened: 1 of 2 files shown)


cmd/internal/ecs/add.go
```diff
@@ -85,20 +85,21 @@ type AddWorkflow struct {
...
-	akitaModificationTagValue = "Akita Software ECS integration"
+	akitaCreationTagKey       = "postman.software:created_by"
+	akitaCreationTagValue     = "Postman Software ECS integration"
+	akitaModificationTagKey   = "postman.software:modified_by"
+	akitaModificationTagValue = "Postman Software ECS integration"
 
 	// Separate AWS secrets for the key ID and key secret
 	// TODO: make these configurable
-	akitaSecretPrefix    = "akita.software/"
+	akitaSecretPrefix    = "postman.software/"
```
mgritter, line 95: Same thing here, let's use either a real domain name or something obviously not a domain name so somebody else can't register it.

cmd/internal/ecs/add.go
```diff
@@ -85,20 +85,21 @@ type AddWorkflow struct {
...
 	defaultKeyIDName     = akitaSecretPrefix + "api_key_id"
 	defaultKeySecretName = akitaSecretPrefix + "api_key_secret"
 
 	// Postman Live Collections Agent image locations
 	akitaECRImage    = "public.ecr.aws/akitasoftware/akita-cli"
 	akitaDockerImage = "akitasoftware/cli"
+	// TODO: postmanDockerImage needed ?
```
mgritter, line 102: No, we aren't using a Dockerhub repository for now (and maybe never.)

cmd/internal/ecs/ecs.go
```diff
@@ -34,7 +34,7 @@ var (
 var Cmd = &cobra.Command{
 	Use:   "ecs",
 	Short: "Add the Postman Live Collections Agent to AWS ECS.",
-	Long:  "The CLI will collect information from you and add the Akita container to an ECS Task.",
+	Long:  "The CLI will collect information from you and add the Postman live agent container to an ECS Task.",
```
mgritter, line 37: Postman Live Collections Agent, I guess, not "live agent".

cmd/internal/ecs/ecs.go
```diff
@@ -63,7 +63,7 @@ var RemoveFromECSCmd = &cobra.Command{
 
 func init() {
 	// TODO: add the ability to specify the credentials directly instead of via an AWS profile?
-	Cmd.PersistentFlags().StringVar(&projectFlag, "project", "", "Your Akita project.")
+	Cmd.PersistentFlags().StringVar(&collectionId, "collection", "", "Your postman collection ID")
```
mgritter, line 66:
```suggestion
	Cmd.PersistentFlags().StringVar(&collectionId, "collection", "", "Your Postman collection ID")
```

cmd/internal/ecs/ecs.go
```diff
@@ -91,40 +91,29 @@ func init() {
...
-			return cmderr.AkitaErr{
-				Err: errors.Wrapf(
-					err,
-					"Could not look up the project %q in the Akita cloud",
-					projectFlag,
-				),
-			}
+		return cmderr.AkitaErr{
+			Err: fmt.Errorf(
+				"could not find the serviceId for %q",
```
mgritter, line 108: This doesn't mean anything to the user; try to avoid showing "serviceID".  There are three cases to consider:

1. The collection ID doesn't exist.
2. The API key presented doesn't have access (or I suppose the API key is absent?)
3. There was an internal problem creating the new service and they need support.

I don't think we can distinguish between 1 + 2, but we should have a friendly error message specifying that the problem is almost certainly their key.  (And if they didn't specify one at all, make clear that that is the problem!)

cmd/internal/ecs/add.go
```diff
@@ -845,7 +846,7 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
 		Name: aws.String("akita-agent"),
 		// TODO: Cpu and Memory should be omitted for Fargate; they take their default values for EC2 if omitted.
 		// For now we can leave the defaults in place, but they might be a bit large for EC2.
-		EntryPoint: []string{"/akita", "apidump", "--project", projectFlag},
+		EntryPoint: []string{"/akita", "apidump", "--collection", collectionId},
 		Environment: []types.KeyValuePair{
 			{Name: aws.String("AKITA_API_KEY_ID"), Value: &apiKey},
 			{Name: aws.String("AKITA_API_KEY_SECRET"), Value: &apiSecret},
```
mgritter, line 852: I assume the replacement by the Postman API Key is a TODO.

cmd/internal/ecs/ecs.go
```diff
@@ -91,40 +91,29 @@ func init() {
...
 	if err != nil {
 		return err
 	}
 
-	// Check project's existence
-	if projectFlag == "" {
-		return errors.New("Must specify the name of your Akita project with the --project flag.")
+	// Check collecton Id's existence
+	if collectionId == "" {
+		return errors.New("Must specify the id of your collection with the --collection flag.")
```
mgritter, line 101:
```suggestion
		return errors.New("Must specify the ID of your collection with the --collection flag.")
```

mgritter (changes requested):
This is a good start.

Most of these are cosmetic; I would like us to do a better job signaling failures to look up the collection ID, though.

cmd/internal/ecs/add.go
```diff
@@ -85,20 +85,21 @@ type AddWorkflow struct {
...
 const (
 	// Tag to use for objects created by the Akita CLI
-	akitaCreationTagKey       = "akita.software:created_by"
-	akitaCreationTagValue     = "Akita Software ECS integration"
-	akitaModificationTagKey   = "akita.software:modified_by"
-	akitaModificationTagValue = "Akita Software ECS integration"
+	akitaCreationTagKey       = "postman.software:created_by"
+	akitaCreationTagValue     = "Postman Software ECS integration"
+	akitaModificationTagKey   = "postman.software:modified_by"
+	akitaModificationTagValue = "Postman Software ECS integration"
```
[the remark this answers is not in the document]
gmann42: done

cmd/internal/ecs/ecs.go
gmann42: Yeah I was thinking of discussing the same.

cmd/internal/ecs/add.go
```diff
@@ -840,15 +840,15 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
...
+		Name: aws.String("postman-lc-agent"),
 		// TODO: Cpu and Memory should be omitted for Fargate; they take their default values for EC2 if omitted.
 		// For now we can leave the defaults in place, but they might be a bit large for EC2.
-		EntryPoint: []string{"/akita", "apidump", "--project", projectFlag},
+		EntryPoint: []string{"/postman-lc-agent", "apidump", "--collection", collectionId},
 		Environment: []types.KeyValuePair{
-			{Name: aws.String("AKITA_API_KEY_ID"), Value: &apiKey},
-			{Name: aws.String("AKITA_API_KEY_SECRET"), Value: &apiSecret},
+			{Name: aws.String("POSTMAN_API_KEY"), Value: &pKey},
+			{Name: aws.String("POSTMAN_ENV"), Value: &pEnv},
```
mgritter, line 851: pEnv may be blank -- in fact that is the expected case for production.

cmd/internal/ecs/ecs.go
gmann42: The correct error states handling is being already done by the function `GetOrCreateServiceIDByPostmanCollectionID`

cmd/internal/ecs/add.go
gmann42: Should we set this with `production` as default value ?

mgritter (approved):

mgritter:
It looks like "Akita" still appears a bunch of times in prompts or log messages, I assume that is TBD in OBS-365?

gmann42:
> It looks like "Akita" still appears a bunch of times in prompts or log messages, I assume that is TBD in OBS-365?

Yeah I tried removing it wherever I saw it. Will scan more closely as a part of OBS-365

PR #242, opened 2023-09-27
[OBS-405]Add setup command for EC2 installation

gmann42:
### Why 
- To onboard as many users as possible
- Current support is available only for Kubernetes, ECS but many of our users are running plain EC2

### What 
- Adds a new command `postman-lc-agent setup` for installation in linux servers

### How
- After user runs `setup` 
- Checks for Postman api key and collectionId existence
- Verifies api key has correct access permission
-  `setupAgentForServer` -> `checkUserPermissions` -> `checkSystemdExists` -> `configureSystemdFiles` -> `enablePostmanAgent` -> `postChecks`

#### func configureSystemdFiles
- Read the env files as go template, fill in the api_key and collection id
- Check or create directories `/etc/default` and `/usr/lib/systemd/system`
- Over-write or create files `/etc/default/postman-lc-agent` and `/usr/lib/systemd/system/postman-lc-agent.service`

### References
- Project Plan: https://postmanlabs.atlassian.net/wiki/spaces/PO/pages/4647125032/Linux+Server+Install
- Ticket: https://postmanlabs.atlassian.net/browse/OBS-405

### Test plan

- [ ] collection Id missing, throw error collection missing
- [ ] POSTMAN_API_KEY missing, throw error api key missing
- [ ] Authorization/collection not existing in backend issues, throw incorrect api key error
- [ ] Non root user, throw error saying use root
- [ ] Systemd doesn't exist, throw error saying not supported yet
**TODO:** Add more test cases


cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,132 @@
+package ec2
+
+import (
+	"context"
+	"os/exec"
+
+	// "github.com/AlecAivazis/survey/v2"
```
mgritter, line 7: Just remove if not needed

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/akitasoftware/akita-cli/cmd/internal/cmderr"
+	"github.com/akitasoftware/akita-cli/printer"
+	"github.com/akitasoftware/akita-cli/telemetry"
+	"github.com/akitasoftware/go-utils/optionals"
+	"github.com/pkg/errors"
+)
+
+// Helper function for reporting telemetry
+func reportStep(stepName string) {
+	telemetry.WorkflowStep("Add to EC2", stepName)
```
mgritter, line 19: I'd call it something like Systemd install?

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,132 @@
...
+	return optionals.None[AddWorkflowState](), nil
+}
+
+func awf_error(err error) (optionals.Optional[AddWorkflowState], error) {
+	return optionals.None[AddWorkflowState](), err
+}
+
+func awf_next(f AddWorkflowState) (optionals.Optional[AddWorkflowState], error) {
+	return optionals.Some[AddWorkflowState](f), nil
+}
```
mgritter, line 38: All this was only necessarily in the interactive version because we allowed people to backup or retry a selection.  In the simplified workflow, there is no retry on error or no prompting, so why keep all this machinery?

cmd/internal/ec2/ec2.go
```diff
@@ -0,0 +1,76 @@
...
+	collectionId string
+
+	// Print out the steps that would be taken, but do not do them
+	dryRunFlag bool
+)
+
+var Cmd = &cobra.Command{
+	Use:          "setup",
+	Short:        "Add the Postman Live Collections Agent to EC2.",
+	Long:         "The CLI will add the Postman Live Collections Agent as a systemd service to your EC2 server.",
```
mgritter, line 25: Although our purpose is to support EC2 users, and that's what we're documenting, I still worry about making it sound like we do something EC2-specific here.  

What we should try to specify is that this enables LCA on the machine they run the command from!  Not some other server, EC2 or not.

So how about replacing "EC2" with "the current server" or "the local server" (although local might not be easy to understand) or "this server"?

cmd/internal/ec2/ec2.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func init() {
+	Cmd.PersistentFlags().StringVar(&collectionId, "collection", "", "Your Postman collection ID")
+	Cmd.MarkPersistentFlagRequired("collection")
+	Cmd.PersistentFlags().BoolVar(
+		&dryRunFlag,
+		"dry-run",
+		false,
+		"Perform a dry run: show what will be done, but do not modify EC2.",
```
mgritter, line 48: EC2 -> systemd?

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,132 @@
...
+	for nextState.IsSome() && err == nil {
+		wf.currentState, _ = nextState.Get()
+		nextState, err = wf.currentState(wf)
+	}
+	if err == nil {
+		telemetry.Success("Add to EC2")
+	} else if errors.Is(err, terminal.InterruptErr) {
+		printer.Infof("Interrupted!\n")
+		telemetry.WorkflowStep("Add to EC2", "User interrupted session")
+		return nil
```
mgritter, line 65: Should be irrelevant if no prompting.

cmd/internal/ec2/ec2.go
mgritter: Or "current instance" or "this instance" -- to use EC2 terminology?

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+	if strings.EqualFold(cu.Name, "root") {
+		return errors.Errorf("Please run the command again with user: root. \n")
+	}
+	return nil
+}
+
+func checkSystemdExists() error {
+	message := "Checking if systemd exists\n"
+	printer.Infof(message)
+	reportStep(message)
```
mgritter, line 82: I think including the \n in the telemetry is probably undesirable.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+	// enable, daemon-reload, start, stop permission for systemctl
+
+	printer.Infof("Checking user permissions \n")
+	cu, err := user.Current()
+	if err != nil {
+		return errors.Errorf("could not get current user. OS returned error: %q \n", err)
+	}
+	if strings.EqualFold(cu.Name, "root") {
+		return errors.Errorf("Please run the command again with user: root. \n")
+	}
```
mgritter, line 75: I think an explanation of what we need to do with root permissions would be appropriate here, like the outline you introduced above.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+func configureSystemdFiles(collectionId string) error {
+	message := "Configuring systemd files\n"
+	printer.Infof(message)
+	reportStep(message)
+
+	// Write collectionId and postman-api-key to go template file
+
+	tmpl, err := template.ParseFS(envFileFS, envFileTemplateName)
+	if err != nil {
+		return errors.Errorf("failed to parse systemd env template file with error :%q\n Please contact observability-support@postman.com  with this log for further assistance.\n", err)
```
mgritter, line 100: Unfortunately, I don't think returning a multi-line error message works very well (but I could be wrong, maybe it looks fine?  Feel free to show some output.)

What I would suggest is reporting the details with glog.Errorf() right away here, and then returning only a short summary error to the top level.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	data := struct {
+		PostmanAPIKey string
+		CollectionId  string
+	}{
```
mgritter, line 106: Please figure out a way to set POSTMAN_ENV as well so that we can easily test against non-production environments.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+		PostmanAPIKey: os.Getenv("POSTMAN_API_KEY"),
+		CollectionId:  collectionId,
+	}
+	envFile, err := os.Create("env-postman-lc-agent")
+	if err != nil {
+		return errors.Errorf("Failed to write values to env file with error %q", err)
```
mgritter, line 112: I wonder if errors.Wrapf() would be more appropriate?

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+	err = tmpl.Execute(envFile, data)
+	if err != nil {
+		return errors.Errorf("Failed to write values to env file with error %q", err)
+	}
+
+	// Ensure /etc/default exists
+	cmd := exec.Command("mkdir", []string{"-p", envFileBasePath}...)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return errors.Errorf("failed to find %s directory with err %q \n", envFileBasePath, err)
```
mgritter, line 124: find -> create

I am not sure whether we should be creating these directories, but if we fail let's be clear what we were attempting.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	data := struct {
+		PostmanAPIKey string
+		CollectionId  string
+	}{
+		PostmanAPIKey: os.Getenv("POSTMAN_API_KEY"),
+		CollectionId:  collectionId,
+	}
+	envFile, err := os.Create("env-postman-lc-agent")
```
mgritter, line 110: I think the temporary files should probably be created in /tmp?  Maybe even with unique names?

Why do we need to create it somewhere other than the final destination?  Is it for atomicity reasons?  A comment would be appropriate.

cmd/internal/ec2/add.go
mgritter: If this is an error message it doesn't need a trailing \n, only log messages do.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,174 @@
...
+func enablePostmanAgent() error {
+	reportStep("Enabling postman-lc-agent as a service")
+
+	cmd := exec.Command("systemctl", []string{"daemon-reload"}...)
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		return errors.Errorf("failed to run systemctl daeomon-reload %q \n", err)
+	}
+	// systemctl start postman-lc-service
+	cmd = exec.Command("systemctl", []string{"start", serviceFileName}...)
```
mgritter, line 159: We need "enable" so that it starts on the next boot.

I think that "enable --now" will also start it immediately

cmd/internal/ec2/add.go
gmann42: We just have to set POSTMAN_ENV in our bash env and in the systemd environment file /etc/default/postman-lc-agent. I've tested this in beta already.

cmd/internal/ec2/add.go
gmann42: My rationale was minimizing possible causes of error. But yeah creating these is a bit questionable.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+	// read/write permission on /usr/lib/system/systemd
+	// enable, daemon-reload, start, stop permission for systemctl
+
+	printer.Infof("Checking user permissions \n")
+	cu, err := user.Current()
+	if err != nil {
+		return errors.Wrapf(err, "could not get current user\n")
+	}
+	if !strings.EqualFold(cu.Name, "root") {
+		printer.Errorf("root user is required to setup systemd service and edit related files")
```
mgritter, line 80: Missing newline:
```suggestion
		printer.Errorf("root user is required to setup systemd service and edit related files.\n")
```

cmd/internal/ec2/ec2.go
```diff
@@ -0,0 +1,76 @@
...
+
+func init() {
+	Cmd.PersistentFlags().StringVar(&collectionId, "collection", "", "Your Postman collection ID")
+	Cmd.MarkPersistentFlagRequired("collection")
+	Cmd.PersistentFlags().BoolVar(
+		&dryRunFlag,
+		"dry-run",
+		false,
+		"Perform a dry run: show what will be done, but do not modify systemd services.",
+	)
```
mgritter, line 49: This flag does not appear to be used?

mgritter (commented):
Is this ready for final review, or still WIP?

gmann42:
> Is this ready for final review, or still WIP?

This is ready for final review.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Exact permissions required are
+	// read/write permissions on /etc/default/postman-lc-agent
+	// read/write permission on /usr/lib/system/systemd
+	// enable, daemon-reload, start, stop permission for systemctl
+
+	printer.Infof("Checking user permissions \n")
+	cu, err := user.Current()
+	if err != nil {
+		return errors.Wrapf(err, "could not get current user\n")
```
mgritter, line 77:
```suggestion
		return errors.Wrapf(err, "could not get current user")
```

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		printer.Errorf("Failed to write values to systemd env file")
+		return err
+	}
+
+	// Ensure /usr/lib/systemd/system exists
+	cmd = exec.Command("mkdir", []string{"-p", serviceFileBasePath}...)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "failed to create %s directory\n", serviceFileBasePath)
```
mgritter, line 142:
```suggestion
		return errors.Wrapf(err, "failed to create %s directory", serviceFileBasePath)
```

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+
+	_, serr := exec.LookPath("systemctl")
+	if serr != nil {
+		printer.Errorf("We don't have support for non-systemd OS as of now.\n For more information please contact observability-support@postman.com.\n")
+		return errors.Errorf("Could not find systemd binary in your OS.")
+	}
+	return nil
+}
+
+func configureSystemdFiles(collectionId string) error {
```
mgritter, line 99: The error messages here are all minimally informational to the end user.  What are they supposed to do next?

Probably, contact the support email for most of them, but are there any that can be corrected?

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+// Starts the postman LCA agent as a systemd service
+func enablePostmanAgent() error {
+	message := "Enabling postman-lc-agent as a service"
+	reportStep(message)
+	printer.Infof(message + "\n")
+
+	cmd := exec.Command("systemctl", []string{"daemon-reload"}...)
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		return err
```
mgritter, line 163: Error needs to be wrapped to signal what we were attempting to do.

cmd/internal/ec2/add.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	// systemctl start postman-lc-service
+	cmd = exec.Command("systemctl", []string{"enable", "--now", serviceFileName}...)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return err
```
mgritter, line 169: Error needs to be wrapped to signal what we were attempting to do.

cmd/internal/ec2/usage_error.go
```diff
@@ -0,0 +1,19 @@
+package ec2
```
mgritter, line 1: UsageError seems to be unused?

mgritter (approved):

cmd/internal/ec2/add.go
gmann42: Yeah, I don't think user can do anything. because lack of systemd is quite unfixable.

PR #243, opened 2023-10-03
OBS-419: Fix distinct ID for Postman API key users.

mgritter:
Previously we were calling /v1/users to get an email to use as a distinct ID only when an Akita API Key was present.

### DIFF
--- a/cfg/credentials.go
+++ b/cfg/credentials.go
@@ -118,3 +118,24 @@ func WritePostmanAPIKeyAndEnvironment(profile, postmanApiKey, postmanEnvironment
 
 	return writeConfigToFile(profile, keyValueMap)
 }
+
+// Check whether credentials are present, of any variety.
+func CredentialsPresent() bool {
+	key, _ := GetPostmanAPIKeyAndEnvironment()
+	if key != "" {
+		return true
+	}
+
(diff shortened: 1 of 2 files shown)


cfg/credentials.go
```diff
@@ -118,3 +118,24 @@ func WritePostmanAPIKeyAndEnvironment(profile, postmanApiKey, postmanEnvironment
...
+	key, secret := GetAPIKeyAndSecret()
+	return key != "" && secret != ""
+}
+
+// If we can't call /v1/user to get a distinct ID, we can try using
+// the credentials provided -- for now this is just an Akita API Key ID.
+// To use a Postman API key we'd have to both obfuscate it (logging
+// the user's API key would be bad) and have a way to map it to a
+// particular user -- seems better to fall back to local IDs.
+func DistinctIDFromCredentials() string {
```
shreys7, line 138: The comment and the implementation seems contradicting 🤔 
We are saying we should not use Postman API Key for distinct id, but we are returning it from this function.

shreys7: Also, shouldn't this function try getting ID from Akita API Key too?

shreys7 (approved):

PR #244, opened 2023-10-20
[OBS-407]Linux agent: Handle reconfiguration

gmann42:
### What
- Handle the case when linux setup command is run again on a system on which it was previously used.
- There can be three possibilities
    - systemd service exists and is enabled: Ask user if old values should be over-written with current input
    - systemd service exists and is disabled: Ask user if old values should be over-written with current input and enable the service
    - systemd service doesn't exist: Proceed with normal flow

Ticket: https://postmanlabs.atlassian.net/browse/OBS-407

### DIFF
--- a/cmd/internal/ec2/add.go
+++ b/cmd/internal/ec2/add.go
@@ -84,6 +84,10 @@ func askToReconfigure() error {
 		},
 		&isReconfigure,
 	)
+	if !isReconfigure {
+		printer.Infof("Exiting setup")
+		return nil
+	}
 	if err != nil {
 		return errors.Wrap(err, "failed to run reconfiguration prompt")
 	}
--- a/cmd/internal/ec2/add.go
+++ b/cmd/internal/ec2/add.go
@@ -41,8 +41,6 @@ var serviceFile string
 //go:embed postman-lc-agent.tmpl
 var envFileFS embed.FS
 
-var isReconfigure string = "n"
-
 // Helper function for reporting telemetry
 func reportStep(stepName string) {
 	telemetry.WorkflowStep("Starting systemd conguration", stepName)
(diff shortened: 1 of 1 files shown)


cmd/internal/ec2/add.go
```diff
@@ -63,6 +72,54 @@ func setupAgentForServer(collectionId string) error {
 	return nil
 }
 
+func askToReconfigure() error {
+	printer.Infof("postman-lc-agent is already present as a systemd service\n")
+	printer.Infof("Helpful commands \n Check status: systemctl status postman-lc-agent \n Disable agent: systemctl disable --now postman-lc-agent \n Check Logs: journalctl -fu postman-lc-agent\n Check env file: cat %s \n Check systemd service file: cat %s \n", envFilePath, serviceFilePath)
+	err := survey.AskOne(
+		&survey.Input{
+			Message: "Over-write old apiKey and collectionId with current values ? (y/n)",
```
mgritter, line 80:
```suggestion
			Message: "Overwrite old API key and Collection ID with current values? (y/n)",
```

If you switch to "Confirm' I don't think the Y/N will be needed.

cmd/internal/ec2/add.go
```diff
@@ -63,6 +72,54 @@ func setupAgentForServer(collectionId string) error {
...
+			Help:    "Enter y(default) or n.",
```
mgritter, line 81: The help should give a longer description of what we're asking them to do.  Something like:

"Should we rewrite the systemd configuration file with the current values of the API key and collection ID?"

I don't know if we can put the specific values here -- parsing the old file to show the change is probably too much work.

I haven't got there yet, the sentence above may not be good because if we're really overwriting then it's not just an update, so the 'help' message or event should make that clear that any customizations made to the file will be discarded.

cmd/internal/ec2/add.go
```diff
@@ -63,6 +72,54 @@ func setupAgentForServer(collectionId string) error {
...
+			Default: "y",
+		},
+		&isReconfigure,
+	)
+	if err != nil {
+		return errors.Wrap(err, "failed to run reconfiguration prompt")
+	}
+	if isReconfigure != "y" {
+		return errors.New("Exiting process because you didn't enter y")
+	}
```
mgritter, line 92: I think this is not an error, and it shouldn't be shown to the user as if it is.

cmd/internal/ec2/add.go
```diff
@@ -63,6 +72,54 @@ func setupAgentForServer(collectionId string) error {
...
+func checkReconfiguration() error {
+
+	cmd := exec.Command("systemctl", []string{"is-enabled", "postman-lc-agent"}...)
+	out, err := cmd.CombinedOutput()
+
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			exitCode := exitError.ExitCode()
+			if exitCode != 1 {
+				return errors.Wrapf(err, "Received non 1 exitcode for systemctl is-enabled")
```
mgritter, line 106: Does this include which exit code it is to help in our own debugging?  (I assume since it's in the original err that it will.)

What does the user see here?  What steps could they take?  If the next step is "contact support with this log output" then we should make that explicit.

cmd/internal/ec2/add.go
```diff
@@ -63,6 +72,54 @@ func setupAgentForServer(collectionId string) error {
...
+				return nil
+			}
+		}
+		return errors.Wrapf(err, "failed to run systemctl is-enabled posman-lc-agent")
+	}
+	if strings.Contains(string(out), enabled) {
+		return askToReconfigure()
+	}
+
+	return errors.Errorf("systemctl is-enabled gave un-supported output:%s", string(out))
```
mgritter, line 120:
```suggestion
	return errors.Errorf("The systemctl is-enabled command produced output this tool doesnt' recognize: %q", string(out))
```

mgritter (changes requested):
Please take another pass over the errors shown to the user and make sure they clearly explain what we were trying to do, and what the user should do next (even if it's only contact us.)

cmd/internal/ec2/add.go
gmann42: It's not possible to easily replicate the systemd error codes, so I don't know what the user will see here. 
We are handling these exit codes
- exit code 0: when service is present and enabled
- exit code 1(service is present but disabled)

Docs: https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#Exit%20status

mgritter (approved):

PR #245, opened 2023-11-10
[POA-155] Capture client/server timeouts as well as Agent parsing errors

shreys7:
In this PR, we have added two sets of changes
- **Capture client and server timeouts**
  - We have used tcpStream's `Accept` function for this purpose. This function gets called on every packet captured from `AssembleWithContext` function of tcpAssembly utility.
  - The stream holds both the flows of the connection, one for the client and one for the server. When we observer a FIN packet for a particular flow, we check if there's data present on the other side of the flow.
  - If there was no packet seen on the other side, this would indicate that one of the sides closed the connection before even receiving any date from the other side, indicating a timeout. 
- **Capture agent parsing errors**
  - When the Agent is not able to decode or parse the request/response body, it only used to log the error.
  - With this PR, we will now also capture the same in the method metadata error field of the witness


trace/backend_collector.go
```diff
@@ -272,6 +276,64 @@ func (c *BackendCollector) processTLSHandshake(tls akinet.TLSHandshakeMetadata)
...
+
+	if !ok {
+		// this should not have happened,a partial witness should have existed for the timeout data passed to us
+		return errors.Errorf("no partial witness found for client timeout %v", witnessID)
+	}
+
+	pair := val.(*witnessWithInfo)
+	if pair.witness.Method.GetMeta() != nil {
+		meta := pair.witness.Method.GetMeta()
+		if meta != nil {
```
mgritter, line 291: You can do this all on one line:

```suggestion
	if meta := pair.witness.Method.GetMeta(); meta != nil {
```

trace/backend_collector.go
```diff
@@ -272,6 +276,64 @@ func (c *BackendCollector) processTLSHandshake(tls akinet.TLSHandshakeMetadata)
...
+			error := pb.HTTPMethodError{
+				Type:    pb.HTTPMethodError_CLIENT_CLOSED,
+				Message: "Client closed the connection",
+			}
+
+			if meta.Errors == nil {
+				meta.Errors = []*pb.HTTPMethodError{&error}
+			} else {
+				meta.Errors = append(meta.Errors, &error)
+			}
```
mgritter, line 301: As above, can append to nil just fine.

trace/backend_collector.go
```diff
@@ -272,6 +276,64 @@ func (c *BackendCollector) processTLSHandshake(tls akinet.TLSHandshakeMetadata)
...
+			error := pb.HTTPMethodError{
+				Type:    pb.HTTPMethodError_SERVER_CLOSED,
+				Message: "Server closed the connection",
+			}
+
+			if meta.Errors == nil {
+				meta.Errors = []*pb.HTTPMethodError{&error}
+			} else {
+				meta.Errors = append(meta.Errors, &error)
+			}
```
mgritter, line 330: Just use append

trace/backend_collector.go
```diff
@@ -272,6 +276,64 @@ func (c *BackendCollector) processTLSHandshake(tls akinet.TLSHandshakeMetadata)
...
+		return errors.Errorf("no partial witness found for client timeout %v", witnessID)
+	}
+
+	pair := val.(*witnessWithInfo)
+	if pair.witness.Method.GetMeta() != nil {
+		meta := pair.witness.Method.GetMeta()
+		if meta != nil {
+			error := pb.HTTPMethodError{
+				Type:    pb.HTTPMethodError_CLIENT_CLOSED,
+				Message: "Client closed the connection",
```
mgritter, line 294: I think the message does not add any context here and could just be omitted.

pcap/stream.go
```diff
@@ -314,6 +333,32 @@ func (c *tcpStream) Accept(tcp *layers.TCP, _ gopacket.CaptureInfo, dir reassemb
 		}
 	}
 
+	// One of the flows initiated a connection close request
+	if tcp.FIN {
```
mgritter, line 337: I do think it is important to handle RST too, or at least leave a FIXME about abnormal termination.

If the server crashes hard (a kernel failure) or there is a load balancer problem directing traffic to the wrong server, then the client will get a RST back instead of a FIN.  Of course, we might not be still around to see the traffic in that case.

The server getting a RST is a bit more obscure.

The comments here do not make it clear how we handle the FIN-ACK from the side that did not initiate the shutdown.

pcap/stream.go
```diff
@@ -55,6 +63,12 @@ type tcpFlow struct {
 	// Context for the FIRST packet that currentParser is processing.
 	currentParserCtx *assemblerCtxWithSeq
 
+	// Indicates if this flow has seen any packet
+	firstPacketSeen bool
+
+	// Indicates who initiated this flow (client or server)
+	initiator *tcpFlowInitiator
```
mgritter, line 70: Consider using go-utils.optional here instead

pcap/stream.go
```diff
@@ -71,9 +85,18 @@ func newTCPFlow(clock clockWrapper, bidiID akinet.TCPBidiID, nf, tf gopacket.Flo
...
 	}
 }
 
+func (f *tcpFlow) FirstPacketSeen() bool {
+	return f.firstPacketSeen
+}
+
+func (f *tcpFlow) TcpFlowInitiator() tcpFlowInitiator {
+	return *f.initiator
+}
```
mgritter, line 98: This is one of the reasons to worry about using *T instead of optional[T], the dereference here is unchecked.

(And might be the nil pointer derefs you are getting in unit tests?  I didn't investigate.)

pcap/stream.go
```diff
@@ -314,6 +359,26 @@ func (c *tcpStream) Accept(tcp *layers.TCP, _ gopacket.CaptureInfo, dir reassemb
...
 	}
 
+	// One of the flows initiated a connection close request
+	if tcp.FIN {
+		if !revFlow.FirstPacketSeen() {
+			// The current flow is the one that initiated the connection close request
+			// and no packets have yet arrived on the reverse flow
+			// this would indicate a timeout on the current side
+			if currFlow.TcpFlowInitiator() == TCP_FLOW_INITIATOR_CLIENT {
+				c.outChan <- currFlow.toPNT(ac.GetCaptureInfo().Timestamp, ac.GetCaptureInfo().Timestamp, akinet.ClientShutdowntMetadata{
```
mgritter, line 369: "Shutdownt" => "Shutdown"?

pcap/stream.go
```diff
@@ -314,6 +359,26 @@ func (c *tcpStream) Accept(tcp *layers.TCP, _ gopacket.CaptureInfo, dir reassemb
 		}
 	}
 
+	// One of the flows initiated a connection close request
+	if tcp.FIN {
+		if !revFlow.FirstPacketSeen() {
+			// The current flow is the one that initiated the connection close request
+			// and no packets have yet arrived on the reverse flow
+			// this would indicate a timeout on the current side
```
mgritter, line 367: I really thing we need to do the design work here of showing the different scenarios and listing out the possibilities to be sure we're covering these correctly.

Client shutdown:
```
client -> SYN
               SYN-ACK <- server
client -> SYN
client -> HTTP request  
               ACK   <- server
              ...
client -> FIN
               FIN-ACK <- server
```

Are we called on every TCP frame, or only those with payloads?  If every TCP frame, then wouldn't revFlow.FirstPacketSeen() be true?  We probably need to set it only when there are bytes in the stream.

Server shutdown:
```
as above until
               ...
               FIN <- server 
client -> FIN-ACK
```

It seems to me that currFlow.TcpFlowInitiator() is not set, because we have not selected a parser.

These are subtle enough that we need unit tests for this code.

mgritter (commented):
Many unit tests appear to be failing, and I'm not yet convinced that the core logic is working yet.  Do you have any test or debugging output that shows otherwise?

PR #246, opened 2023-11-15
[POA-41] Update analytics functions to support amplitude functions

mudit-postman:
* Replace Segment related keys and mentioned with Amplitude Keys and configs
* Updated few event names to make them consistent in Amplitude

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -73,7 +73,7 @@ func Init(isLoggingEnabled bool) {
 	}
 	if amplitudeKey == "" {
 		if isLoggingEnabled {
-			printer.Infof("Telemetry unavailable; no Segment key configured.\n")
+			printer.Infof("Telemetry unavailable; no Amplitude key configured.\n")
 			printer.Infof("This is caused by building from source rather than using an official build.\n")
 		}
 		analyticsClient = nullClient{}
@@ -91,13 +91,15 @@ func Init(isLoggingEnabled bool) {
 				Build:     version.GitVersion(),
 				Namespace: "",
 			},
-			// No output from the Segment library
+			// No output from the Amplitude library
 			IsLoggingEnabled: false,
+			// Add agent prefix to events before sending to Amplitude
+			IsAgent: true,
 		},
 	)
 	if err != nil {
 		if isLoggingEnabled {
-			printer.Infof("Telemetry unavailable; error setting up Segment client: %v\n", err)
+			printer.Infof("Telemetry unavailable; error setting up Analytics(Amplitude) client: %v\n", err)
 			printer.Infof("Postman support will not be able to see any errors you encounter.\n")
 			printer.Infof("Please send this log message to observability-support@postman.com.\n")
 		}
@@ -112,7 +114,7 @@ func getDistinctID() string {
 	// Otherwise use the configured API Key.
 	// Failing that, try to use the user name and host name?
 
-	id := os.Getenv("AKITA_SEGMENT_DISTINCT_ID")
+	id := os.Getenv("POSTMAN_ANALYTICS_DISTINCT_ID")
 	if id != "" {
 		return id
 	}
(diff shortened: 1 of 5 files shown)


telemetry/telemetry.go
```diff
@@ -114,7 +114,7 @@ func getDistinctID() string {
 	// Otherwise use the configured API Key.
 	// Failing that, try to use the user name and host name?
 
-	id := os.Getenv("AKITA_SEGMENT_DISTINCT_ID")
+	id := os.Getenv("POSTMAN_ANALYTICS_DISTINCT_ID")
```
mudit-postman, line 117: Do we need to do any changes for this in documentations?
cc @mgritter @liujed

telemetry/telemetry.go
```diff
@@ -24,7 +24,7 @@ var (
 	analyticsEnabled bool
 
 	// Client key; set at link-time with -X flag
```
mgritter, line 26: Needs a corresponding change in superstar/cli-postman and superstar/cli-akita

telemetry/telemetry.go
```diff
@@ -83,23 +83,23 @@ func Init(isLoggingEnabled bool) {
...
 				Name:      "akita-cli",
 				Version:   version.ReleaseVersion().String(),
 				Build:     version.GitVersion(),
 				Namespace: "",
 			},
-			// No output from the Segment library
+			// No output from the Amplitude library
 			IsLoggingEnabled: false,
-			// IsMixpanelEnabled: false,  -- irrelevant for us, leaving at default value
-			BatchSize: 1, // disable batching
```
mgritter, line 97: I don't recall precisely why I disabled batching, but I think it had to do with catching errors during command-line execution, or maybe other abnormal exits.  I think that should still be part of our use of Amplitude, if batching can be disabled?

telemetry/telemetry.go
mgritter: No, this is not a documented feature.  The ability to disable analytics is more important.

telemetry/telemetry.go
```diff
@@ -255,10 +255,10 @@ func APIError(method string, path string, e error) {
 	)
 }
 
-// Report a failure withoout a specific error object
+// Report a failure without a specific error object
 func Failure(message string) {
 	analyticsClient.Track(distinctID(),
-		message,
+		fmt.Sprintf("Unknown Error: %s", message),
```
mgritter, line 261: I am unsure about the need to add the prefixes as you have done here.  I don't think we really felt their absence using Mixpanel, and the "type" field could be used to filter.  Is this an Amplitude or Postman best practice that we should understand?

We should not use "Unknown Error" here, though, because the failures can be known errors that simply don't come with a Go "error".

telemetry/telemetry.go
```diff
@@ -268,7 +268,7 @@ func Failure(message string) {
 // Report success of an operation
 func Success(message string) {
 	analyticsClient.Track(distinctID(),
-		message,
+		fmt.Sprintf("Success in %s", message),
```
mgritter, line 271: For example, in this case we have only two instances:

"Success in rotate learn session"
"Success in Add to ECS"

which is inconsistent capitalization :) and I don't think the success adds much?

mgritter (approved):

telemetry/telemetry.go
mudit-postman: We can disable batching we can use `FlushQueueSize` config for this.

telemetry/telemetry.go
mudit-postman: I have just added these prefixes to follow a convention with the event names, also it can help us in the easier resolution of new events that might be added in the Amplitude.
Amplitude doesn't automatically start tracking a new event, we need to tell Amplitude to start tracking the event. Like in the attached image, events that are not being tracked are marked as Unexpected we need to add them to the list to make them live. 
https://app.amplitude.com/data/postman/Postman/events/POA-41-migrate-to-amplitude/latest?view=All
image

mudit-postman:
Fixed `backend_collector_test.go` test case.
We have updated the akita-libs version here which has changes related to obfuscation [PR](https://github.com/akitasoftware/akita-libs/pull/196), due to which test case was failing

PR #251, opened 2024-02-16
[POA-731] Integrate new service info and user info endpoints

mudit-postman:
* Integrated two new endpoints in akita-cli
* Currently they are not in use, will be added in further PRs

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -125,9 +125,11 @@ func VerifyServiceByServiceID(c rest.FrontClient, serviceID akid.ServiceID) erro
 		}
 
 		if httpErr.StatusCode == 404 {
-			return fmt.Errorf("there is not service with given ID %s. Ensure that your service ID is correct", serviceID)
+			//lint:ignore ST1005 This is a user-facing error message
+			return fmt.Errorf("There is no service with given ID %s. Ensure that your service ID is correct", serviceID)
 		} else if httpErr.StatusCode == 403 {
-			return fmt.Errorf("you cannot send traffic to the service with ID %s. "+
+			//lint:ignore ST1005 This is a user-facing error message
+			return fmt.Errorf("You cannot send traffic to the service with ID %s. "+
 				"Ensure that your service ID is correct and that you have required permissions. "+
 				"If you do not have required permissions, please contact the workspace administrator", serviceID)
 		}
--- a/util/util.go
+++ b/util/util.go
@@ -102,7 +102,7 @@ func VerifyServiceByServiceID(c rest.FrontClient, serviceID akid.ServiceID) erro
 	failedToVerifyServiceErrMsg := "Failed to verify service for given serviceID: %s\n"
 
 	// Check if service is already verified and cached
-	if service, found := serviceIDCache.Get(akid.String(serviceID)); found {
+	if service, found := serviceIDCache.Get(serviceID.String()); found {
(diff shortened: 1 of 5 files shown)


util/util.go
```diff
@@ -94,6 +96,50 @@ func GetServiceIDByName(c rest.FrontClient, name string) (akid.ServiceID, error)
...
+		var errorResponse rest.ErrorResponse
+		if err := json.Unmarshal(httpErr.Body, &errorResponse); err != nil {
+			printer.Stderr.Debugf(failedToVerifyServiceErrMsg, err)
+			return errors.Wrap(err, unexpectedErrMsg)
+		}
+
+		if httpErr.StatusCode == 404 {
+			return fmt.Errorf("there is not service with given ID %s. Ensure that your service ID is correct", serviceID)
+		} else if httpErr.StatusCode == 403 {
+			return fmt.Errorf("you cannot send traffic to the service with ID %s. "+
```
mgritter, line 130: Blocking: It's not clear to me without digging deeper how these will be displayed.  However, it seems inconsistent that the errors messages on line 100 and 102 are capitalized, while these two are not.  So we should resolve that to ensure the error is displayed in a way that looks good to the user.

```suggestion
			return fmt.Errorf("There is no service with given ID %s. Ensure that your service ID is correct", serviceID)
		} else if httpErr.StatusCode == 403 {
			return fmt.Errorf("You cannot send traffic to the service with ID %s. "+
```

I don't think "service" and "service ID" are the terms we are using, can you confirm with our documentation of what we are calling the Insights projects?  In user-facing messages we should ensure we're following consistent naming -- even though it will be some work to work through all the places where we say "services" in the CLI right now, let's not introduce new ones.

util/util.go
```diff
@@ -94,6 +96,50 @@ func GetServiceIDByName(c rest.FrontClient, name string) (akid.ServiceID, error)
 	return akid.ServiceID{}, errors.Errorf("cannot determine project ID for %s", name)
 }
 
+func VerifyServiceByServiceID(c rest.FrontClient, serviceID akid.ServiceID) error {
+	unexpectedErrMsg := "Something went wrong while starting the Agent. " +
+		"Please contact Postman support (observability-support@postman.com) with the error details"
+	failedToVerifyServiceErrMsg := "Failed to verify service for given serviceID: %s\n"
+
+	// Check if service is already verified and cached
+	if service, found := serviceIDCache.Get(akid.String(serviceID)); found {
```
mgritter, line 105: Non-blocking: Blarg, is the cast still necessary?  Should we have a typed cache in go-utils?

How about `serviceID.String()` instead? (and on line 138)

mgritter (approved):
Conditionally approved, please ensure error messages are consistent and use the appropriate user-facing vocabulary.

util/util.go
mudit-postman: > it seems inconsistent that the errors messages on line 100 and 102 are capitalized, while these two are not

Have not capitalized them due to the lint warning, added lint ignore.

util/util.go
mudit-postman: Not sure; I thought to use `serviceID.String()` initially, but then noticed that we had used `akid.String()` in other places also. So, decided to proceed with this.

liujed: `akid.String` is old. `String()` is preferable.

PR #252, opened 2024-02-16
[POA-731] Update project flag in cli commands

mudit-postman:
* Tried to update --project work with minimal breaking change
  * Updated `apidump` and `ecs add` commands
* Will be removing the deprecated codes in further PRs

### DIFF
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -151,8 +151,9 @@ type Args struct {
 type apidump struct {
 	*Args
 
-	backendSvc  akid.ServiceID
-	learnClient rest.LearnClient
+	backendSvc     akid.ServiceID
+	backendSvcName string
+	learnClient    rest.LearnClient
 
 	startTime   time.Time
 	dumpSummary *Summary
@@ -169,7 +170,7 @@ func newSession(args *Args) *apidump {
 
 // Is the target the Akita backend as expected, or a local HAR file?
 func (a *apidump) TargetIsRemote() bool {
-	return a.Out.AkitaURI != nil || a.PostmanCollectionID != ""
+	return a.Out.AkitaURI != nil || a.PostmanCollectionID != "" || a.ServiceID != akid.ServiceID{}
 }
 
 // Lookup the service and create a learn client targeting it.
@@ -186,13 +187,15 @@ func (a *apidump) LookupService() error {
 		}
 
 		a.backendSvc = backendSvc
+		a.backendSvcName = "Postman_Collection_" + a.PostmanCollectionID
 	} else {
-		err := util.VerifyServiceByServiceID(frontClient, a.ServiceID)
+		serviceName, err := util.GetServiceNameByServiceID(frontClient, a.ServiceID)
 		if err != nil {
 			return err
 		}
 
 		a.backendSvc = a.ServiceID
+		a.backendSvcName = serviceName
 	}
 
 	a.learnClient = rest.NewLearnClient(a.Domain, a.ClientID, a.backendSvc)
@@ -556,11 +559,10 @@ func (a *apidump) Run() error {
 				return errors.Errorf("Cannot automatically rotate sessions when a session name is provided.")
 			}
 		}
-	} else if args.PostmanCollectionID != "" {
+	} else if (args.PostmanCollectionID != "" || args.ServiceID != akid.ServiceID{}) {
 		args.Out.AkitaURI = &akiuri.URI{
-			ObjectType: akiuri.TRACE.Ptr(),
-			// Placeholder name for postman collectionIds
-			ServiceName: "Postman_" + args.PostmanCollectionID,
+			ObjectType:  akiuri.TRACE.Ptr(),
+			ServiceName: a.backendSvcName,
 			ObjectName:  util.RandomLearnSessionName(),
 		}
 	}
--- a/apidump/apidump.go
+++ b/apidump/apidump.go
@@ -88,6 +88,9 @@ type Args struct {
 	// Args used to using agent with Postman
 	PostmanCollectionID string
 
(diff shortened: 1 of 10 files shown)


cmd/internal/apidump/apidump.go
```diff
@@ -67,18 +68,18 @@ var Cmd = &cobra.Command{
...
+		// If --project was given, convert serviceFlag to serviceID.
 		if serviceFlag != "" {
-			uri, err := akiuri.Parse(akiuri.Scheme + serviceFlag)
+			parsedID, err := akid.ParseID(serviceFlag)
 			if err != nil {
-				return errors.Wrap(err, "bad project name")
+				return errors.Wrap(err, "failed to parse service ID")
 			}
-			outFlag.AkitaURI = &uri
+			serviceID = parsedID.(akid.ServiceID)
```
liujed, line 82: This should use `akid.ParseIDAs` instead.

cmd/internal/apidump/apidump.go
```diff
@@ -106,22 +107,18 @@ var Cmd = &cobra.Command{
 		}
 
 		// Allow specification of an alternate rotation time, default 1h.
+		// We can rotate the trace if we're sending the output to a cloud-based trace.
 		// But, if the trace name is explicitly given, or selected by tag,
 		// or we're sending the output to a local file, then we cannot rotate.
 		traceRotateInterval := time.Duration(0)
-		if outFlag.AkitaURI != nil {
+		if (outFlag.AkitaURI != nil && outFlag.AkitaURI.ObjectName == "") || projectID != "" || postmanCollectionID != "" {
```
mudit-postman, line 114: Combine separate checks for `outFlag.AkitaURI` and `outFlag.AkitaURI.ObjectName`, since we won't be using this flag now. So no need to throw a separate error.
The main aim is to handle `collectionId` and `projectId` for trace rotation

apidump/apidump.go
```diff
@@ -553,11 +559,10 @@ func (a *apidump) Run() error {
...
 			}
 		}
-	} else if args.PostmanCollectionID != "" {
+	} else if (args.PostmanCollectionID != "" || args.ServiceID != akid.ServiceID{}) {
 		args.Out.AkitaURI = &akiuri.URI{
-			ObjectType: akiuri.TRACE.Ptr(),
-			// Placeholder name for postman collectionIds
-			ServiceName: "Postman_" + args.PostmanCollectionID,
+			ObjectType:  akiuri.TRACE.Ptr(),
+			ServiceName: a.backendSvcName,
```
mudit-postman, line 565: Create a new `backendSvcName` var since we will be getting the project name from BE in case of `--project`.
Setting this var above in `LookupService()` function

cmd/internal/apidump/apidump.go
```diff
@@ -192,40 +190,20 @@ var Cmd = &cobra.Command{
 }
 
 func init() {
-	Cmd.Flags().Var(
```
mudit-postman, line 195: Removed `--out`, `--service` and `--cluster` commands.

Can't completely remove the `outFlag` var since it might need some major refactoring. outFlag will be `nil` now.

cmd/internal/ecs/add.go
```diff
@@ -848,9 +848,17 @@ func modifyTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowS
 		}...)
 	}
 
+	var entryPoint []string
+
+	if collectionId != "" {
```
mudit-postman, line 853: Currently supporting both `--collection` and `--project` flags

versilis (approved):
Looks good

apidump/apidump.go
liujed: **Blocker.** How does this get used? AkitaURIs don't uniquely identify a service anymore, now that service names are no longer unique.

cmd/internal/apidump/apidump.go
```diff
@@ -192,40 +189,20 @@ var Cmd = &cobra.Command{
...
-		"Your Akita project. Exactly one of --out or --project must be specified.")
-	Cmd.Flags().MarkDeprecated("project", "For use by Akita users")
+		"Your Postman Insights projectID.")
 
 	Cmd.Flags().StringVar(
 		&postmanCollectionID,
 		"collection",
 		"",
-		"Your Postman collectionID.")
+		"Your Postman collectionID. Exactly one of --project, --collection must be specified.")
```
liujed, line 202: Why are we still supporting `--collection`? One of our stated goals is to have users only reinstall once. Won't those that continue to use `--collection` need to reinstall again once we drop support for this in the back end?

liujed (changes requested):
I have concerns here about our continued internal use of AkitaURIs that warrant blocking this PR: https://github.com/akitasoftware/akita-cli/pull/252#discussion_r1501204555.

Not necessarily looking for a code change: if you think this won't be a problem, it would be enough to explain why.

apidump/apidump.go
mudit-postman: Yes, serviceName is not being used now. The only use is in the log message.
https://github.com/akitasoftware/akita-cli/blob/80cebed4d07a9a28ba84cb9c5f2154c484817e6e/apidump/apidump.go#L588

Example: `[INFO] Created new trace on Postman Cloud: akita://apidump-native:trace:ribbon-carp-a9e625e9`

I have added it just to maintain consistency with the old code. 
When using `--collections` we don't have any service name so we use `Postman_random()`, and now since we have serviceName, I am setting it. Also, it can be useful in logs.

The best option is to completely remove the `out` variable and `akita.Uri` concept, which requires more change, which we can do during Akita Sunset.

liujed (approved):

PR #253, opened 2024-02-23
[POA-752] Mark old commands as deprecated

mudit-postman:
* Marked old commands as deprecated, they will not show in help text now
* `might be removed` for `ec2` and `kube`
* `will be removed` for `daemon` and `nginx`


mudit-postman:
Do we need to make these commands, unusable? Like error out like, `apispec`
https://github.com/akitasoftware/akita-cli/blob/ee18ae86fbe2d48b4483575eda7080205e5ef731/cmd/internal/apispec/cmd.go#L17

liujed (approved):
This can go as-is.

But thinking on this some more, how about if we removed the "will be removed" commands altogether, but leave their supporting code? Maybe also give the same treatment to commands that are already errors. I'm hoping this would be an easy blast-away-code operation. If you need to do surgery/refactoring of any kind to achieve this, then let's just go with what is here.

mudit-postman:
> This can go as-is.
> 
> But thinking on this some more, how about if we removed the "will be removed" commands altogether, but leave their supporting code? Maybe also give the same treatment to commands that are already errors. I'm hoping this would be an easy blast-away-code operation. If you need to do surgery/refactoring of any kind to achieve this, then let's just go with what is here.

Removed old deprecated commands from `root` Cobra Command, but have kept their related code.

PR #258, opened 2024-03-08
POA-894 Remove mentions of Akita (+ other fixes & cleanup)

liujed:
* Reinstated the Easter egg.
* Fixed the command for stopping the agent in ECS.
* Removed code for commands that have been removed. Keeping `nginx`, however.
* Removed dead `apidiff` code.
* Removed `ci_guard`. No longer relevant, now that we're not a Github app anymore.
* Removed support for deployments.
* Don't create `.akita` directory.
* `ecs` command: stop checking for Akita container images.
* Rename environment variables: `AKITA_` → `POSTMAN_`.
* `kube` command: removed support for Akita `--project` flag and Akita authentication.
* Removed support for Akita back-end domains.
* Removed mentions of Akita.


liujed:
If it helps, each commit can be reviewed in isolation.

cmd/internal/ecs/add.go
```diff
@@ -533,7 +533,7 @@ func getTaskState(wf *AddWorkflow) (nextState optionals.Optional[AddWorkflowStat
 	// Check that the postman-insights-agent is not already present
 	for _, container := range output.ContainerDefinitions {
 		image := aws.ToString(container.Image)
-		if matchesImage(image, postmanECRImage) || matchesImage(image, akitaECRImage) || matchesImage(image, akitaDockerImage) {
+		if matchesImage(image, postmanECRImage) {
```
mgritter, line 536: I think we should probably continue looking for the old docker images?  We don't want users who convert over to Postman to have two copies of the agent running.

cmd/internal/kube/inject.go
```diff
@@ -41,15 +38,15 @@ var injectCmd = &cobra.Command{
 	Short: "Inject the Postman Insights Agent into a Kubernetes deployment",
 	Long:  "Inject the Postman Insights Agent into a Kubernetes deployment or set of deployments, and output the result to stdout or a file",
 	RunE: func(_ *cobra.Command, args []string) error {
-		if projectNameFlag == "" && postmanCollectionID == "" {
+		if postmanCollectionID == "" {
 			return cmderr.AkitaErr{
-				Err: errors.New("--collection must be specified. Or, if you are an Akita user, use --project instead."),
+				Err: errors.New("--collection must be specified."),
 			}
 		}
```
mgritter, line 45: This seems like it is left in an odd state where we still require a collection ID, not a project ID?

mgritter (approved):
Looks good.

This closes https://postmanlabs.atlassian.net/browse/POA-103 as well, I think?

cmd/internal/kube/inject.go
liujed: Yeah, converting this to a project ID was a larger change, and therefore out of scope, since we don't support Kubernetes currently. At least this moves us in the right direction by dropping the conflicting `--project` flag that was used by Akita users to identify an Akita project by name.

cmd/internal/ecs/add.go
liujed: In b4a9481.

PR #259, opened 2024-03-08
[POA-885] Replace serviceID with projectID

mudit-postman:
* We are showing serviceID in logs only in the `util.go` functions while looking for service.
* Replaced service -> project and serviceID -> projectID
* Also done changes in  `GetServiceIDByPostmanCollectionID()` since we are not removing it now itself.


mgritter (approved):
Looks OK; I couldn't find references elsewhere in the akita-cli code other than this file, either.

PR #262, opened 2024-04-10
[POA-877|669] Refactor telemetry events

mudit-postman:
* Refactored `distinctID()` function in `telemetry.go` to have both `user_id` and `team_id` instead of only `distinct_id`.
* We will be sending both `user_id` and `team_id` as base properties along with other event specific properties
* Done changes in the event names to combine various events with a single generic event, each separate event can be identified by the properties

More info on events refactoring: https://docs.google.com/spreadsheets/d/19u7kHBunBLcbtfQjTvrrnQ_cgiOOEKYO9lhWviuCuk8/edit?usp=sharing

### DIFF
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -123,6 +123,8 @@ func doInit() {
 		return
 	}
 
+	analyticsEnabled = true
+
 	userID, teamID, err = getUserIdentity() // Initialize user ID and team ID
 	if err != nil {
 		if isLoggingEnabled {
@@ -137,8 +139,6 @@ func doInit() {
 	// Set up automatic reporting of all API errors
 	// (rest can't call telemetry directly because we call rest above!)
 	rest.SetAPIErrorHandler(APIError)
-
-	analyticsEnabled = true
 }
 
 func getUserIdentity() (string, string, error) {
--- a/telemetry/telemetry.go
+++ b/telemetry/telemetry.go
@@ -15,6 +15,7 @@ import (
 	"github.com/akitasoftware/akita-cli/rest"
 	"github.com/akitasoftware/akita-cli/version"
 	"github.com/akitasoftware/akita-libs/analytics"
+	"github.com/akitasoftware/go-utils/maps"
 )
 
 var (
@@ -27,10 +28,11 @@ var (
 	// Client key; set at link-time with -X flag
 	defaultAmplitudeKey = ""
 
-	// Store the distinct ID; run through the process
+	// Store the user ID and team ID; run through the process
 	// of getting it only once.
-	userDistinctID   string
-	userDistinctOnce sync.Once
+	userID           string
+	teamID           string
+	userIdentityOnce sync.Once
 
 	// Timeout talking to API.
 	// Shorter than normal because we don't want the CLI to be slow.
@@ -112,14 +114,15 @@ func Init(isLoggingEnabled bool) {
 	}
 }
 
-func getDistinctID() string {
-	// If we have a user email, use that!
+func getUserIdentity() (string, string) {
+	// If we can get user details use userID and teamID
 	// Otherwise use the configured API Key.
-	// Failing that, try to use the user name and host name?
+	// Failing that, try to use the user name and host name.
+	// In latter 2 cases teamID will be empty.
 
 	id := os.Getenv("POSTMAN_ANALYTICS_DISTINCT_ID")
 	if id != "" {
-		return id
+		return id, ""
 	}
 
 	// If there's no credentials configured, skip the API call and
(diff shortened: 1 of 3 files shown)


telemetry/telemetry.go
```diff
@@ -143,44 +145,43 @@ func getDistinctID() string {
 	}
 
 	// Try to derive a distinct ID from the credentials, if present, even
-	// if the /v1/user call failed.
+	// if the getUser() call failed.
 	keyID := cfg.DistinctIDFromCredentials()
 	if keyID != "" {
-		return keyID
+		return keyID, ""
```
mgritter, line 151: Maybe use "unknown" if we are sending this string as-is?  I don't see a check for "" in the use site, and it feels like dealing with the empty string would be difficult in Amplitude?

telemetry/telemetry.go
```diff
@@ -143,44 +145,43 @@ func getDistinctID() string {
...
 	}
 
 	localUser, err := user.Current()
 	if err != nil {
-		return "unknown"
+		return "", ""
```
mgritter, line 156: Same here; how useful is "" as a user ID?

telemetry/telemetry.go
```diff
@@ -133,8 +136,7 @@ func getDistinctID() string {
 		frontClient := rest.NewFrontClient(rest.Domain, GetClientID())
 		userResponse, err := frontClient.GetUser(ctx)
 		if err == nil {
-			// User postman user's userID as distinctID for telemetry
-			return fmt.Sprint(userResponse.ID)
+			return fmt.Sprint(userResponse.ID), fmt.Sprint(userResponse.TeamID)
```
mgritter, line 139: Line 132 still mentions email.

mgritter (changes requested):

mudit-postman:
Disable telemetry if we cannot get any userID. Also don't send team_id if it is empty string.

telemetry/telemetry.go
```diff
@@ -321,3 +321,20 @@ func Shutdown() {
 		printer.Infof("Please send the CLI output to %s.\n", consts.SupportEmail)
 	}
 }
+
+// Attempts to track an event using the provided event name and properties.
+// It initializes the user identity, adds the user ID and team ID to the event properties,
+// and then sends the event to the analytics client.
```
mgritter, line 327: I don't think it "initializes the user identity"?

telemetry/telemetry.go
```diff
@@ -27,10 +28,11 @@ var (
...
 	defaultAmplitudeKey = ""
 
-	// Store the distinct ID; run through the process
+	// Store the user ID and team ID; run through the process
 	// of getting it only once.
-	userDistinctID   string
-	userDistinctOnce sync.Once
+	userID           string
+	teamID           string
+	userIdentityOnce sync.Once
```
mgritter, line 35: I can't find this sync.Once used anywhere in this PR.

Do you believe it is no longer necessary (because Init is only invoked once)?  I do not have a high degree of confidence that telemetry.Blah() will never be called before Init() and would like to have something sensible happen. Using the nullClient{} is probably fine?

Probably we can just delete this, but please do think about synchronization issues.  The Go memory model does allow us to write to analyticsClient (and userID, and teamID) before starting any other goroutines that use them.  Is this sufficient (as I evidently believed in a previous version) or do we need to do explicit locking?

mudit-postman: * I have reused `sync.Once` to ensure the client is initiated before sending the events.
  * There is one minor issue related to the init process logging. If tomorrow we add a new `telemetry.Init(false)` call and somehow `telemetry.tryTrackingEvent()` is called before this, we will show logs.
* Does this seem okay, or should we explore the `chan`? This could lock the `tryTrackingEvent()` function's routine

telemetry/telemetry.go
```diff
@@ -51,9 +58,14 @@ func (_ nullClient) Close() error {
 	return nil
 }
 
+func Init(loggingEnabled bool) {
+	isLoggingEnabled = loggingEnabled
+	initClientOnce.Do(doInit)
+}
+
 // Initialize the telemetry client.
 // This should be called once at startup either from the root command or from a subcommand that overrides the default PersistentPreRun.
```
mgritter, line 67: Blocking: When you refactor, please move the comment to the correct location.

telemetry/telemetry.go
```diff
@@ -321,3 +331,22 @@ func Shutdown() {
 		printer.Infof("Please send the CLI output to %s.\n", consts.SupportEmail)
 	}
 }
+
+// Attempts to track an event using the provided event name and properties. It adds the user ID
+// and team ID to the event properties, and then sends the event to the analytics client.
+// If there is an error sending the event, a warning message is printed.
+func tryTrackingEvent(eventName string, eventProperties maps.Map[string, any]) {
+	// precondition: analyticsClient is initialized
+	Init(true)
```
mgritter, line 340: The old code could log in the case where getDistinctID() was called before Init(), so it's no worse than before.

I don't love that you're providing an explicit value here that updates a global variable each time.

Why not call 
```
	initClientOnce.Do(doInit)
```
directly?

mgritter (approved):

PR #263, opened 2024-04-23
POA-114 Add ecs cf-fragment and ecs task-def subcommands

liujed:
Added `ecs cf-fragment` subcommand for printing an AWS CloudFormation fragment for adding the agent to ECS as a sidecar.

Added `ecs task-def` subcommand for printing a task definition that can be added to an ECS cluster to run the agent as a daemon in host-networking mode on every EC2 instance in the cluster.

Removed `--collection` from `ecs` subcommands.

### DIFF
--- a/cmd/internal/ecs/ecs.go
+++ b/cmd/internal/ecs/ecs.go
@@ -36,6 +36,9 @@ var (
 	// Location of credentials file.
 	awsCredentialsFlag string
 
+	// Output in YAML instead of JSON.
+	yamlFlag bool
+
 	// Print out the steps that would be taken, but do not do them
 	dryRunFlag bool
 )
@@ -106,6 +109,13 @@ func init() {
 		"dry-run",
 		false,
 		"Perform a dry run: show what will be done, but do not modify ECS.",
+	)
+
+	PrintCloudFormationFragmentCmd.Flags().BoolVar(
+		&yamlFlag,
+		"yaml",
+		false,
+		"Output as YAML instead of JSON",
 	)
 
 	// Support for credentials in a nonstandard location
@@ -181,12 +191,16 @@ func printCloudFormationFragment(cmd *cobra.Command, args []string) error {
 		isEssential,
 	)
 
-	result, err := ecs_cloudformation_utils.ContainerDefinitionToJSONForCloudformation(agentContainer)
+	formatter := ecs_cloudformation_utils.ContainerDefinitionToJSONForCloudFormation
+	if yamlFlag {
+		formatter = ecs_cloudformation_utils.ContainerDefinitionToYAMLForCloudFormation
+	}
+	result, err := formatter(agentContainer)
 	if err != nil {
-		return errors.Wrapf(err, "unable to format container definition as JSON")
+		return errors.Wrapf(err, "unable to format CloudFormation fragment")
 	}
 
-	fmt.Println(string(result))
+	fmt.Println(result)
 	return nil
 }
 
--- a/cmd/internal/ecs/ecs.go
+++ b/cmd/internal/ecs/ecs.go
@@ -10,7 +10,7 @@ import (
 	"github.com/akitasoftware/akita-cli/util"
 	"github.com/akitasoftware/akita-libs/akid"
 	"github.com/akitasoftware/go-utils/optionals"
-	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/akitasoftware/go-utils/slices"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


liujed:
Sample interaction for both commands. This has NOT been tested in AWS.

```
$ POSTMAN_API_KEY=PMAK-XXXXX postman-insights-agent ecs cf-fragment --project svc_YYYYY
[INFO] Telemetry unavailable; no Amplitude key configured.
[INFO] This is caused by building from source rather than using an official build.
[INFO] Postman Insights Agent 0.0.0
{
  "Command": null,
  "Cpu": 0,
  "DependsOn": null,
  "DisableNetworking": null,
  "DnsSearchDomains": null,
  "DnsServers": null,
  "DockerLabels": null,
  "DockerSecurityOptions": null,
  "EntryPoint": [
    "/postman-insights-agent",
    "apidump",
    "--project",
    "svc_YYYYY"
  ],
  "Environment": [
    {
      "Name": "POSTMAN_API_KEY",
      "Value": "PMAK-XXXXX"
    }
  ],
  "EnvironmentFiles": null,
  "Essential": false,
  "ExtraHosts": null,
  "FirelensConfiguration": null,
  "HealthCheck": null,
  "Hostname": null,
  "Image": "docker.postman.com/postman-insights-agent",
  "Interactive": null,
  "Links": null,
  "LinuxParameters": null,
  "LogConfiguration": null,
  "Memory": null,
  "MemoryReservation": null,
  "MountPoints": null,
  "Name": "postman-insights-agent",
  "PortMappings": null,
  "Privileged": null,
  "PseudoTerminal": null,
  "ReadonlyRootFilesystem": null,
  "RepositoryCredentials": null,
  "ResourceRequirements": null,
  "Secrets": null,
  "StartTimeout": null,
  "StopTimeout": null,
  "SystemControls": null,
  "Ulimits": null,
  "User": null,
  "VolumesFrom": null,
  "WorkingDirectory": null
}

$ POSTMAN_API_KEY=PMAK-XXXXX bin/postman-insights-agent ecs task-def --project svc_YYYYY
[INFO] Telemetry unavailable; no Amplitude key configured.
[INFO] This is caused by building from source rather than using an official build.
[INFO] Postman Insights Agent 0.0.0
{
  "Compatibilities": [
    "EC2"
  ],
  "ContainerDefinitions": [
    {
      "Command": null,
      "Cpu": 0,
      "DependsOn": null,
      "DisableNetworking": null,
      "DnsSearchDomains": null,
      "DnsServers": null,
      "DockerLabels": null,
      "DockerSecurityOptions": null,
      "EntryPoint": [
        "/postman-insights-agent",
        "apidump",
        "--project",
        "svc_YYYYY"
      ],
      "Environment": [
        {
          "Name": "POSTMAN_API_KEY",
          "Value": "PMAK-XXXXX"
        }
      ],
      "EnvironmentFiles": null,
      "Essential": true,
      "ExtraHosts": null,
      "FirelensConfiguration": null,
      "HealthCheck": null,
      "Hostname": null,
      "Image": "docker.postman.com/postman-insights-agent",
      "Interactive": null,
      "Links": null,
      "LinuxParameters": null,
      "LogConfiguration": null,
      "Memory": null,
      "MemoryReservation": null,
      "MountPoints": null,
      "Name": "postman-insights-agent",
      "PortMappings": null,
      "Privileged": null,
      "PseudoTerminal": null,
      "ReadonlyRootFilesystem": null,
      "RepositoryCredentials": null,
      "ResourceRequirements": null,
      "Secrets": null,
      "StartTimeout": null,
      "StopTimeout": null,
      "SystemControls": null,
      "Ulimits": null,
      "User": null,
      "VolumesFrom": null,
      "WorkingDirectory": null
    }
  ],
  "Cpu": "256",
  "DeregisteredAt": null,
  "EphemeralStorage": null,
  "ExecutionRoleArn": null,
  "Family": "postman-insights-agent",
  "InferenceAccelerators": null,
  "IpcMode": "",
  "Memory": "512",
  "NetworkMode": "host",
  "PidMode": "",
  "PlacementConstraints": null,
  "ProxyConfiguration": null,
  "RegisteredAt": null,
  "RegisteredBy": null,
  "RequiresAttributes": null,
  "RequiresCompatibilities": [
    "EC2"
  ],
  "Revision": 0,
  "RuntimePlatform": {
    "CpuArchitecture": "X86_64",
    "OperatingSystemFamily": "LINUX"
  },
  "Status": "ACTIVE",
  "TaskDefinitionArn": null,
  "TaskRoleArn": null,
  "Volumes": null
}
```

cmd/internal/ecs/ecs.go
```diff
@@ -93,17 +112,21 @@ func init() {
...
 	_, err := cmderr.RequirePostmanAPICredentials("The Postman Insights Agent must have an API key in order to capture traces.")
 	if err != nil {
 		return err
 	}
 
-	// Check collecton Id's existence
+	// Check that a collection or project is provided.
 	if collectionId == "" && projectId == "" {
 		return errors.New("exactly one of --project or --collection must be specified")
 	}
```
mgritter, line 132: I'm not clear why we need to continue to support --collection in the ecs subcommands.  Seems like we could deprecate them as long as it is still supported in 'apidump' for those who haven't modified their tasks yet.

cmd/internal/ecs/ecs.go
```diff
@@ -127,9 +150,107 @@ func addAgentToECS(cmd *cobra.Command, args []string) error {
...
+		},
+		"family":      "postman-insights-agent",
+		"networkMode": types.NetworkModeHost,
+		"requiresCompatibilities": []types.Compatibility{
+			types.CompatibilityEc2,
+		},
+		"cpu":    "512",
+		"memory": "512",
+		"runtimePlatform": map[string]any{
+			"cpuArchitecture":       types.CPUArchitectureX8664,
```
mgritter, line 244: We have successfully run Graviton, but I don't think that is in scope for this project.

cmd/internal/ecs/ecs.go
```diff
@@ -127,9 +150,107 @@ func addAgentToECS(cmd *cobra.Command, args []string) error {
...
+		optionals.None[string](),
+		optionals.None[string](),
+		isEssential,
+	)
+
+	// XXX Whereas the AWS SDK defines a `types.TaskDefinition`, that type is
+	// intended for use in calling the AWS APIs, not for printing JSON fragments
+	// for pasting into the AWS console. We therefore take the brittle approach of
+	// manually cobbling together an object that will serialize to the appropriate
+	// JSON value.
```
mgritter, line 210: Wow, yuck. The Task definition SDK's JSON does not match the JSON format accepted in the console?  I wondered whether there was some better option out there; Terraform, naturally, just uses the SDK, and the CDK uses CloudFormation format.

I think it would be better to define appropriate structs here (if we can't get them elsewhere), but it is non-blocking.

cmd/internal/ecs/ecs.go
```diff
@@ -127,9 +150,107 @@ func addAgentToECS(cmd *cobra.Command, args []string) error {
...
+
+	const isEssential = false
+	agentContainer := makeAgentContainerDefinition(
+		optionals.None[string](),
+		optionals.None[string](),
+		optionals.None[string](),
+		isEssential,
+	)
+
+	result, err := json.MarshalIndent(agentContainer, "", "  ")
```
mgritter, line 183: I can't recall here, do we need to support YAML output?

CDK is a mixed bag.  The artifacts created during normal operation (as temporary files) seem to be JSON, but `cdk synthesize` returns YAML, and I think the diffs were YAML too.

cmd/internal/ecs/ecs.go
```diff
@@ -64,6 +69,20 @@ var RemoveFromECSCmd = &cobra.Command{
...
+	Short: "Print an AWS CloudFormation fragment for adding the Postman Insights Agent to AWS ECS.",
+	Long:  "Print a code fragment that can be inserted into a CloudFormation template to add the Postman Insights Agent as a sidecar container.",
+	RunE:  printCloudFormationFragment,
+}
+
+var PrintECSTaskDefinitionCmd = &cobra.Command{
+	Use:   "task-def",
+	Short: "Print an AWS ECS task definition for running the Postman Insights Agent in daemon mode.",
+	Long:  "Print a task definition that can be added to an ECS cluster to run the Postman Insights Agent as a daemon in host-networking mode on every EC2 instance in the cluster.",
+	RunE:  printECSTaskDefinition,
```
mgritter, line 83: I wonder whether these subcommands should be set up to call telemetry.Init(false) to disable logging from the telemetry system -- do they otherwise log?  I think for the K8S injector we tried to make sure it did not, but the use case is not quite the same.

mgritter (approved):
Potentially blocking: does CF need to be in Yaml?  Or have a YAML option?

Other than that, looks OK.

cmd/internal/ecs/ecs.go
liujed: Commit ec735e8 defines types in a separate package for both the AWS console and CloudFront.

cmd/internal/ecs/ecs.go
liujed: The three lines below are the only things extra that is printed. Only the first two lines are from the telemetry system, and won't be there for the release build and are to stderr. I've modified the third to be printed to stderr. Everything else is on stdout.
```
[INFO] Telemetry unavailable; no Amplitude key configured.
[INFO] This is caused by building from source rather than using an official build.
[INFO] Postman Insights Agent 0.0.0
```

cmd/internal/ecs/ecs.go
liujed: CloudFormation accepts both. Added `--yaml` to `ecs cf-fragment` in 421d986.

cmd/internal/ecs/ecs.go
liujed: Removed in f346cb7.

PR #265, opened 2024-05-06
[POA-527] Add apidump flags support in ecs command

mudit-postman:
Plan: https://postmanlabs.atlassian.net/wiki/spaces/PO/pages/5053218962/Plan+Insights+Agent+ECS+and+other+tasks+for+supporting+50+users

* Added support for the following eight flags: `filter`, `host-allow`, `host-exclusions`, `interfaces`, `path-allow`, `path-exclusions`, `rate-limit` and `stats-log-delay`.
* Added new vars and persistent flags for them
* Used these flags in `makeAgentContainerDefinition()` in `add.go` while constructing the entryPoint slice.
* These flags can also be used in all 3 sub-commands, i.e. `add`, `cf-fragment` and `task-def`

---

Reviewers:
* Primary: @shreys7 
* Secondary: @mgritter

### DIFF
--- a/cmd/internal/ecs/add.go
+++ b/cmd/internal/ecs/add.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -943,11 +944,29 @@ func makeAgentContainerDefinition(
 	addOptToEnv("POSTMAN_ECS_SERVICE", ecsService)
 	addOptToEnv("POSTMAN_ECS_TASK", ecsTaskDefinitionFamily)
 
+	// Pass apidump flags as it is, the apidump command will parse them.
+	// We are already handling the default values in apidump command
 	entryPoint := []string{
 		"/postman-insights-agent",
 		"apidump",
 		"--project",
 		projectId,
+		"--filter",
+		filterFlag,
+		"--host-allowlist",
+		strings.Join(hostAllowlistFlag, ","),
+		"--host-exclusions",
+		strings.Join(hostExclusionsFlag, ","),
+		"--interfaces",
+		strings.Join(interfacesFlag, ","),
+		"--path-allowlist",
+		strings.Join(pathAllowlistFlag, ","),
+		"--path-exclusions",
+		strings.Join(pathExclusionsFlag, ","),
+		"--rate-limit",
+		strconv.FormatFloat(rateLimitFlag, 'f', -1, 64),
+		"--stats-log-delay",
+		strconv.Itoa(statsLogDelay),
 	}
 
 	// XXX If we instantiate any new fields in the container definition here, we
--- a/cmd/internal/ecs/ecs.go
+++ b/cmd/internal/ecs/ecs.go
@@ -3,6 +3,7 @@ package ecs
 import (
 	"fmt"
 
+	"github.com/akitasoftware/akita-cli/apispec"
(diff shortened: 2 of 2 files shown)


shreys7 (approved):

shreys7:
How are we planning to test this @mudit-postman ?

mudit-postman:
> How are we planning to test this @mudit-postman ?
Testing this by logging the flag values and using postman beta env's keys. Also I am commenting the final step where the updated definition is applied.

cmd/internal/ecs/ecs.go
```diff
@@ -115,6 +127,16 @@ func init() {
...
 
+	// initialize apidump flags as persistent flags for the ecs command
+	Cmd.PersistentFlags().StringVar(&filterFlag, "filter", "", "Used to match packets going to and coming from your API service.")
+	Cmd.PersistentFlags().StringSliceVar(&hostAllowlistFlag, "host-allow", nil, "Allows only HTTP hosts matching regular expressions.")
+	Cmd.PersistentFlags().StringSliceVar(&hostExclusionsFlag, "host-exclusions", nil, "Removes HTTP hosts matching regular expressions.")
+	Cmd.PersistentFlags().StringSliceVar(&interfacesFlag, "interfaces", nil, "List of network interfaces to listen on. Defaults to all interfaces on host.")
+	Cmd.PersistentFlags().StringSliceVar(&pathAllowlistFlag, "path-allow", nil, "Allows only HTTP paths matching regular expressions.")
+	Cmd.PersistentFlags().StringSliceVar(&pathExclusionsFlag, "path-exclusions", nil, "Removes HTTP paths matching regular expressions.")
+	Cmd.PersistentFlags().Float64Var(&rateLimitFlag, "rate-limit", apispec.DefaultRateLimit, "Number of requests per minute to capture.")
+	Cmd.PersistentFlags().IntVar(&statsLogDelay, "stats-log-delay", apispec.DefaultStatsLogDelay_seconds, "Print packet capture statistics after N seconds.")
```
mgritter, line 138: Omit if undocumented or hidden.

Possibly, make hidden -- this is not something we would normally expect somebody to do when running noninteractively.

cmd/internal/ecs/add.go
```diff
@@ -943,11 +944,38 @@ func makeAgentContainerDefinition(
...
 		"/postman-insights-agent",
 		"apidump",
 		"--project",
 		projectId,
+		"--filter",
+		filterFlag,
+		"--rate-limit",
+		strconv.FormatFloat(rateLimitFlag, 'f', -1, 64),
+		"--stats-log-delay",
+		strconv.Itoa(statsLogDelay),
```
mgritter, line 959: I think we should avoid adding new "always present" parameters like this.  Here's my reasoning:

1) It clutters up the command line, and we already have people asking us to take the command line via environment variables or config file.  Shouldn't give the impression that there is a lot of information here.

2) --rate-limit in particular is something where we may want to switch to a much different approach (a global rate limit shared among agents) and having it present for every agent install makes that transition harder.

3) Changes to the default behavior should be applied to old apidump installation when they upgrade to a new agent version.

mgritter (changes requested):

cmd/internal/ecs/ecs.go
mudit-postman: Instead of marking them as hidden, we should document these flags in LC docs.
Users can use these flags while running `ecs add` for their ECS tasks

mgritter (approved):

PR #267, opened 2024-05-20
[POA-168] Clean up authentication messages

mudit-postman:
In this PR, we have updated some of the error messages shown to users in case of errors.
* Remove mention of login command with postman-insights-agent
* Remove mention of contacting the postman support team if a user is using --collection flag
* The 3rd message mentioned in the JIRA ticket has already been fixed in earlier releases.
* Don't show the raw response body to the user for external API call errors; instead, show a generic message for unhandled cases.

**PS**: To show the user the proper message, we will need to follow a standard error response model, which will involve changes in both the Agent and Superstar services. Otherwise, we will face issues in mapping the response to the proper Golang object.

### DIFF
--- a/cmd/internal/cmderr/checks.go
+++ b/cmd/internal/cmderr/checks.go
@@ -17,10 +17,10 @@ func RequirePostmanAPICredentials(explanation string) (string, error) {
 		if env.InDocker() {
 			printer.Infof("Please set the POSTMAN_API_KEY environment variables on the Docker command line.\n")
 		} else {
-			printer.Infof("Use the POSTMAN_API_KEY environment variables, or run 'postman-insights-agent login'.\n")
+			printer.Infof("Please set the POSTMAN_API_KEY environment variables, either in your shell session or prepend it to postman-insights-agent command.\n")
 		}
-
-		return "", AkitaErr{Err: errors.New("could not find an Postman API key to use")}
+		//lint:ignore ST1005 This is a user-facing error message
+		return "", AkitaErr{Err: errors.New("Could not find an Postman API key to use")}
 	}
 
 	return key, nil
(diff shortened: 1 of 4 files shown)


shreys7 (changes requested):
Some nitpicks, rest it looks good.

util/util.go
```diff
@@ -173,8 +173,7 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, ctx context.Context,
 func GetOrCreateServiceIDByPostmanCollectionID(c rest.FrontClient, collectionID string) (akid.ServiceID, error) {
 	// Normalize the collectionID.
 	collectionID = strings.ToLower(collectionID)
-	unexpectedErrMsg := "Something went wrong while starting the Agent. " +
-		"Please contact Postman support (" + consts.SupportEmail + ") with the error details"
```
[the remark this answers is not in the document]
mudit-postman: It is mentioned in ticket to not mentioned this for `--collection` flag which is deprecated now.

shreys7 (approved):

cmd/internal/cmderr/checks.go
```diff
@@ -17,10 +17,10 @@ func RequirePostmanAPICredentials(explanation string) (string, error) {
 		if env.InDocker() {
 			printer.Infof("Please set the POSTMAN_API_KEY environment variables on the Docker command line.\n")
 		} else {
-			printer.Infof("Use the POSTMAN_API_KEY environment variables, or run 'postman-insights-agent login'.\n")
+			printer.Infof("Please set the POSTMAN_API_KEY environment variables, either in your shell session or prepend it to postman-insights-agent command.\n")
```
mgritter, line 20:
```suggestion
			printer.Infof("Please set the POSTMAN_API_KEY environment variable, either in your shell session or prepend it to postman-insights-agent command.\n")
```

I am unsure whether the part about prepending it is clear (though I know what you mean?)  Maybe we should assume people know how to set environment variables?

util/util.go
```diff
@@ -173,8 +173,7 @@ func GetServiceIDByPostmanCollectionID(c rest.FrontClient, ctx context.Context,
...
+	unexpectedErrMsg := "Something went wrong while starting the agent."
```
mgritter, line 176: This is fine, because the collection path is deprecated.

However, the point I was trying to make in the bug is that if the problem is easily fixable by the user we should recommend such a fix first, not recommend contacting Postman support immediately.

mgritter (approved):

